PR #1, opened 2026-05-13
chore: import codebase

itsjoeoui:
Imports the codebase from the existing private repo as a single squashed commit. This is the initial code drop for the open-source `instro` repo.


hoehner (approved):
Confirmed this is the expected file set.

PR #8, opened 2026-05-23
feat(cli): add `instro doctor` environment health check

hoehner:
Migrated from nominal-io/instrumentation#224.


hoehner:
Revisiting this at a later date.

PR #19, opened 2026-05-26
refactor(daq)!: remove InstroDAQFacade; driver owns channel/timing state

nhschwab:
Tracks #28 / CON-2488

Drop APIInstroDAQ, InstroDAQFacade, and DAQDriverBase.hal. Drivers now hold the single source of truth for configured channels and timing config; InstroDAQ exposes that state to users through @property proxies.

- DAQDriverBase.__init__ initializes ai/ao/di/do/relay_channels dicts, all four *_hw_timing_config slots, and points_in_buffer. Concrete drivers call super().__init__() and populate the dicts inside their configure_* methods. define_relay_channel's default now records on self.relay_channels.
- InstroDAQ drops its parallel dicts. ai_channels, ao_channels, di_channels, do_channels, relay_channels, the *_hw_timing_config attrs, and an aggregate `channels` list are now @property delegations into self._driver. A new `daq.driver` accessor exposes the driver itself.
- All four vendor drivers (Keysight 34980A, NI DAQmx, MCC, LabJack T-series) call super().__init__() and record each configure_* call on the matching dict, so the proxy path always sees up-to-date state.
- tests/daq/test_daq_drivers.py: add _make_mock_driver() helper that pre-initializes the dicts and wires configure_* side_effects to populate them, matching real-driver contract through the new proxy path.
- AGENTS.md: new "DAQ driver state tracking" section with the four authoring rules (super().__init__(), populate dicts in configure_*, read via self.<dict>, no InstroDAQ reach-back). Patterns bullet about DAQDriverBase notes it's the one base with a default __init__.
- docs/guides/instrumentation/daq.mdx: rewrite "Driver-owned state" to describe base-initialized dicts + @property proxies; new code snippet shows the @property delegation shape.
- migrate_to_instro.py + v1.mdx: NominalDAQFacade joins the removed-facade passthrough list so users hit a clear ImportError.

## Summary

## Type of change

- [ ] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [x] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

I ran the scripts at `examples/daq` with a LabJack T8, NI USB-6002, MCC USB-1616HS-4, and a Keysight 34980A and verified the test results in Nominal Core workbooks. Screenshots of those workbooks for each device are included below:

### LabJack T8

image

image

image

image

image

image

image

### NI USB-6002

image

image

image

image

image

image

image

### MCC USB-1616HS-4

image

image

image

image

image

### Keysight 34980A

image

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers


instro/daq/daq.py
hoehner, line 112: We should add an @abc.abstactmethod here to enforce override of this method like the others in this class.

hoehner: Or we should raise a "NotImplemented" error by default if something hasn't overridden this.

instro/daq/drivers/keysight_34980a.py
```diff
@@ -172,10 +175,10 @@ def fetch_analog(
         self,
     ) -> KeysightData:
         """Block until the buffer holds at least one full per-channel batch, then drain a channel-aligned chunk."""
-        channels = self.hal.ai_channels
-        hw_timing_config = self.hal.ai_hw_timing_configs
-        num_channels = len(channels)
-        min_points_per_fetch = hw_timing_config.samples_per_channel * num_channels
+        if self.ai_hw_timing_config is None:
+            raise RuntimeError("configure_ai_hw_timing() must be called before fetch_analog().")
```
hoehner: I think a user might be confused by this error message because the DAQ instrument doesn't expose this specific "configure_ai_hw_timing" method - it just exposes "configure_ai_sample_rate" which calls the "configure_ai_hw_timing" method on the driver.

hoehner: Lets check the other drives as well to ensure we throw the right error to the user.

hoehner (changes requested):
Two minor comments

instro/daq/daq.py
nhschwab: Addressed

instro/daq/drivers/keysight_34980a.py
nhschwab: Addressed

hoehner (dismissed):

hoehner (approved):

PR #31, opened 2026-05-28
feat(psu): add ovp, ocp, and remote sense method signatures

wharwood:
## Summary

Creates the framework to allow INSTRO-2 to be closed. This extends the existing PSU api to include OVP, OCP, and remote sense features.

## Type of change

- [ ] Bug fix (`fix`)
- [X] New feature (`feat`)
- [x] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Added tests to ensure base implementation raises NotImplementedError.

## Tests

- [ ] No tests — explain why: This is a protocol update, not a concrete implementation.

## Checklist

## Notes for reviewers


instro/i2c/i2c.py
```diff
@@ -9,9 +9,9 @@
 from typing import Literal, Optional
 
 from instro.i2c.types import CommandDevice, RegisterDevice, SystemDefinition
-from instro.utils import Command, Instrument, Measurement
-from instro.utils.instrument import publish_command, publish_measurement
-from instro.utils.publishers.publisher import Publisher
+from instro.lib import Command, Instrument, Measurement
+from instro.lib.instrument import publish_command, publish_measurement
+from instro.lib.publishers.publisher import Publisher
```
hoehner: Let's figure out how to fix this double import. I know you didn't add it but I've been trying to sniff these out as we march towards open source.

hoehner: Its okay for this to exist in some of the __init__.py files but I'd like to eliminate them in files like this one.

instro/lib/instrument.py
```diff
@@ -10,12 +10,12 @@
...
 
-from instro.utils.publishers import NominalCorePublisher
-from instro.utils.publishers.channel_buffer import (
+from instro.lib.publishers import NominalCorePublisher
+from instro.lib.publishers.channel_buffer import (
     DequeInMemoryPublisher,
 )
-from instro.utils.publishers.publisher import Publisher
-from instro.utils.types import BackgroundDaemonConfig, Command, Measurement
+from instro.lib.publishers.publisher import Publisher
```
hoehner: Another good candidate to try and remove the double import

instro/i2c/i2c.py
hoehner: Looking at this closer it seems like maybe we just need to re-export the "Publisher" class in this file. 
instro/lib/publishers/__init__.py

hoehner:
One other thought, should we have explicit FeatureNotSupported errors in each of the child drivers or in the top level Base driver? The former has us duplicate a lot of code. Not the end of the world but just something to consider.

hoehner:
As for the "docs check" this just means you need to regenerate examples with "just gen-examples" and then commit the results.

hoehner (changes requested):

instro/i2c/i2c.py
wharwood: These should be fixed.

instro/lib/instrument.py
wharwood: Is this more what you were thinking?

`
from instro.lib.publishers import NominalCorePublisher, Publisher
from instro.lib.publishers.channel_buffer import (
    DequeInMemoryPublisher,
)
`

hoehner (dismissed):

wharwood:
@hoehner I added more methods to better prep for concrete implementation. I also did the base class error default you asked for.

hoehner (approved):

PR #41, opened 2026-05-29
feat(daq)!: require fully-qualified NI physical channel names

### LINKED ISSUE #40: NI DAQmx driver should require fully-qualified physical channel names
## Problem

The NI DAQmx driver (`NIDAQDriver`) concatenates its `device_id` constructor argument onto the `physical_channel` you pass when configuring a channel (e.g. `f"{self._device_id}/{channel.physical_channel}"`). This forces a device-relative channel form (`ai0`, `port0/line0`) and prevents expressing the full DAQmx name. It is also inconsistent with the other DAQ drivers (LabJack, MCC, Keysight), which all pass `physical_channel` through verbatim.

## Proposal

Require the fully-qualified name in `physical_channel` (e.g. `Dev1/ai0`, `Dev1/port0/line0`) and pass it to NI-DAQmx verbatim. Keep `device_id` on the constructor; it is still used for `open()` device verification and DAQmx task naming.

## Scope

- Drop the `device_id` prefix in `configure_ai_channel`, `configure_ao_channel`, `configure_do_channel`, `configure_di_channel`.
- Fix `define_digital_channel`: discriminate port vs line on `/line` rather than any `/` (a fully-qualified port like `Dev1/port0` now contains a `/`).
- Update DAQ examples and the `daq.mdx` channel-naming guidance to fully-qualified NI names.
- Add a note to the v1 migration guide (separate axis from the `legacy_naming` published-name change; no compatibility flag).
- Add NI driver unit tests (none existed) covering verbatim pass-through and the port/line validation.

## Breaking change

Yes. NI users must prefix their `physical_channel` strings with the device name. Channel aliases and published names are unchanged.

hoehner:
Closes #40

## What

`NIDAQDriver` no longer concatenates `device_id` onto `physical_channel`. The value now reaches NI-DAQmx verbatim, so callers supply the fully-qualified name (`Dev1/ai0`, `Dev1/port0/line0`) instead of the device-relative form (`ai0`, `port0/line0`). This brings NI in line with the LabJack, MCC, and Keysight drivers, which already pass `physical_channel` through unchanged.

## Changes

- **Driver** (`packages/instro-daq-ni/.../nidaq.py`): drop the `device_id` prefix in `configure_ai_channel`, `configure_ao_channel`, `configure_do_channel`, `configure_di_channel`. Fix `define_digital_channel` to discriminate port vs line on `/line` rather than any `/`, since a fully-qualified port (`Dev1/port0`) now contains a `/`.
- **Examples**: NI channel constants in `examples/daq/*.py` updated to fully-qualified names; regenerated the affected Mintlify example pages.
- **Docs**: `daq.mdx` channel-naming notes now show fully-qualified NI names; added an `NI physical channel naming` section to the v1 migration guide (a separate axis from the `legacy_naming` published-name change, with no compatibility flag).
- **Tests** (new `tests/daq/ni/test_nidaq_driver.py`): there were no NI driver tests. Added mocked-`nidaqmx` coverage for verbatim AI/AO/DO/DI pass-through plus the port/line validation.

`device_id` stays on the constructor; the driver still uses it for `open()` device verification and DAQmx task naming. Channel aliases and published names are unchanged.

## Breaking change

NI users must prefix `physical_channel` strings with the device name. Documented in the v1 migration guide.

## Verification

`just check` clean (ruff format, mypy, ruff lint); `tests/daq` (39).


docs/guides/migration/v1.mdx
```diff
@@ -303,6 +305,28 @@ class InstroMyThing(Instrument):
 
 If you previously copied the old `_read_to_command` / `_read_to_measurement` helper pattern into your subclass, delete those copies. The base class methods replace them, and the decorators enforce the channel-naming contract at runtime.
 
+## NI physical channel naming
```
smithednominal, line 308: I might link here: https://www.ni.com/docs/en-US/bundle/ni-daqmx/page/physchannames.html#GUID-7CB1537B-6D48-41A9-BF7E-5510456B6CB4
it would also be good to clarify that ranges, lists are not supported or are handled internal to instro (assuming they are not because these are all single channels in the examples)

Benefit is it points direct to NI's docs and so it might help people figure out channel names on their own if they're having issues (I've certainly been on systems where someone renamed the channels in MAX for example).

smithednominal: oh, may be off: looks like ports are supported but not ai ranges?

smithednominal (dismissed):
Most of the changes make sense. Suggestions:

1- I suggested adding a link to NI documentation in v1.mdx and also clarifying there (and perhaps elsewhere) that we don't for example support lists (dev1/ai0:ai2)

2- One thing I note is that define_digital_channel checks to make sure if you want a line you get a line and if you want a port you get a port. you might want to add a similar check in the AI code for list or comma-separated syntax?

3- I didn't know this existed, but since in "open" you are checking against `system.devices` I wonder if it wouldn't be a good idea to:
(a) add to the raised RuntimeError information about supported devices. Ie "Device X not found; Connected devices include [A, B, C]"
(b) Adding to the DAQDriverBase a method for listing available channels at runtime. However I'd not be too surprised if labjack for example doesnt even have a way to list these, so maybe its biting off more than we should try to chew.

I'm marking approved because I don't see these as blockers, but I think they are worth considering

docs/guides/migration/v1.mdx
hoehner: Correct, we don't support ai ranges today but we do support ports.

hoehner: I do like the clarification and docs reference idea. Adding those.

smithednominal (approved):
Updates lgtm, no more notes

PR #54, opened 2026-06-01
fix(psu): expect "+0" from TDK Lambda SYST:ERR? response

### LINKED ISSUE #53: bug: Fix expected response from TDK Lambda Genesys SYS:ERR? query
### What happened?

Fix TDK Lambda error response formatting
Is expecting "0" in response to SYS:ERR?
Should be expecting "+0" in response to SYS:ERR?

### What did you expect to happen?

Expected error parsing to be handled correctly.

### How can we reproduce it?

```python
run "psu.py" against a TDK power supply.
```

### instro version

0.6

### Python version

3.13

### Operating system

macOS

### Device (if relevant)

_No response_

### Logs or output

```shell

```

### Before submitting

- [x] I've searched existing issues and didn't find a duplicate.

hoehner:
Closes #53.

## What

The TDK Lambda Genesys `SYSTEM:ERROR?` query returns a signed status code (e.g. `+0,"No error"`), but `_check_errors` only accepted a bare `0` prefix. Every healthy response started with `+0`, so it failed the `startswith("0")` check and was misreported as an error.

## Change

- `tdk_lambda_genesys.py`: match the `+0` prefix, consistent with the Siglent driver.
- `tests/psu/test_psu_drivers.py`: update the TDK fixtures/cases to the real `+0,"No error"` device response (they previously encoded the buggy `0,...` form).

## Testing

- `just check` (format + mypy + lint) on the changed files
- `uv run pytest tests/psu/test_psu_drivers.py` — 77 passed

### DIFF
--- a/instro/psu/drivers/tdk_lambda_genesys.py
+++ b/instro/psu/drivers/tdk_lambda_genesys.py
@@ -50,5 +50,5 @@ class TDKLambdaGenesys(PSUDriverBase):
 
     def _check_errors(self) -> None:
         err = self._visa.query("SYSTEM:ERROR?")
-        if not err.startswith("0"):
+        if not err.startswith("+0"):
             raise RuntimeError(f"TDK Lambda PSU reported error: {err}")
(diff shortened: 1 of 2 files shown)


wharwood (approved):
LGTM

PR #68, opened 2026-06-08
feat(ethernetip): batch reads

benliepert:
Migrated from nominal-io/instrumentation#211.


packages/instro-unstable/instro/unstable/ethernetip/ethernetip.py
```diff
@@ -114,8 +117,9 @@ def __init__(
             self.start()
 
     def _define_background_daemon(self) -> None:
-        for tag in self._config.polled_tags:
-            self.add_background_daemon_function(self.read_tag, tag.alias)
+        self._polled_tags = self._config.polled_tags
+        if self._polled_tags:
+            self.add_background_daemon_function(self._poll_batched_tags)
```
NthTensor, line 122: I have some questions relating to the lifetime of this background things, but they don't relate directly to the content of this PR. I see that the other backends are also using `add_background_daemon_function`.

packages/instro-unstable/instro/unstable/ethernetip/ethernetip.py
```diff
@@ -178,6 +182,33 @@ def read_tag(self, alias: str, **kwargs) -> Measurement:
...
         )
 
+    def _poll_batched_tags(self, **kwargs) -> Measurement | None:
+        """Read every polled tag in one batched request."""
+        if not self._polled_tags:
+            return None
+
+        tag_by_name = {tag.tag_name: tag for tag in self._polled_tags}
+        results = self._read_tags_raw([tag.tag_name for tag in self._polled_tags])
+        timestamp = time.time_ns()
```
NthTensor, line 192: Why is this being sampled in python and not in rust?

NthTensor (approved):
I think both of my comments are larger questions that I will discuss with @hoehner.

packages/instro-unstable/instro/unstable/ethernetip/ethernetip.py
benliepert: my initial thought was "ethernet IP doesn't give us a native way to get the _real_ timestamp, so best-effort at the python side is fine"
[INSTRO-349 ](https://linear.app/nominal-io/issue/INSTRO-349/timestamp-values-rust-side) to track

hoehner (approved):

PR #75, opened 2026-06-09
chore(ethernetip): docs

benliepert:
Migrated from nominal-io/instrumentation#227.


docs/guides/instrumentation/protocols/ethernetip.mdx
```diff
@@ -0,0 +1,595 @@
...
+| Reject out-of-range setpoints before PLC writes | `write_min` / `write_max` on numeric tags |
+| Reach a PLC CPU behind a local chassis/backplane | `connection.route_path.hops` with `type: "backplane"` |
+| Read or write a string tag manually | `EtherNetIpSession.read_tag()` / `EtherNetIpSession.write_tag()` |
+| Read several PLC tags in one native request | `EtherNetIpSession.read_tags([...])` |
+
+## Key concepts
+
+### Lifecycle
+
+The EtherNet/IP workflow has six moves:
```
NthTensor: "moves" is a bit of an odd word choice here.

docs/guides/instrumentation/protocols/ethernetip.mdx
```diff
@@ -0,0 +1,595 @@
...
+
+1. Define PLC tags in a JSON config file or an `EtherNetIPConfig`.
+2. Instantiate `EtherNetIPDevice(config)` with an optional connection override.
+3. Call `open()` to establish the native EtherNet/IP session.
+4. Call `start()` to begin background polling when timing is configured.
+5. Read or write tags by alias with type validation.
+6. Call `close()` to close the session and stop background polling.
```
NthTensor, line 149: What happens if I fail to call `close()`?

benliepert: The daemon thread will live for the duration of the python process, publishers won't be closed (so they may hold resources, as is the case with the Core publisher but not the Connect publisher), and underlying drivers won't close (which may hold resources as well)

I'll add a note that ideally you use instruments via a `with` statement, which handles calling `close()` for you when the statement exits

NthTensor (dismissed):

cartercanedy (dismissed):
LGTM

Note: verify the documented package extra name for EtherNet/IP before merging. The docs currently refer to `instro-unstable[ethernetip]`; confirm that matches the intended published package metadata and release state.

cartercanedy (approved):

PR #76, opened 2026-06-09
chore(ethernetip): examples

benliepert:
Migrated from nominal-io/instrumentation#220.


docs/reference/mkdocs.yml
```diff
@@ -34,6 +34,47 @@ nav:
   - Protocols:
       - protocols/index.md
       - Modbus: protocols/modbus.md
+  - Examples:
+      - examples/index.md
+      - DAQ:
+          - examples/daq/daq_read_analog_sw_timed.md
```
cartercanedy: mkdocs is emitting a warning about a missing DAQ example: daq_multi_rate_ni.md is generated by generate_examples.py but isn't listed in this nav block, so it's reported as an omitted file.

docs/reference/mkdocs.yml
```diff
@@ -34,6 +34,47 @@ nav:
   - Protocols:
       - protocols/index.md
       - Modbus: protocols/modbus.md
```
cartercanedy, line 36: Should we also include some EtherNet/IP stuff in the mkdocs configuration? Modbus has a protocol reference page here, but EtherNet/IP only got example pages -- there's no equivalent reference entry for the instro.unstable.ethernetip API.

cartercanedy: this might be noise if the idea is to revisit once ethernet/ip graduates from `unstable`

cartercanedy (approved):

PR #79, opened 2026-06-09
fix(lib): stop closing the shared pyvisa ResourceManager in VisaDriver

### LINKED ISSUE #78: VisaDriver.close() closes the shared pyvisa ResourceManager, killing other instrument sessions
## Problem

`VisaDriver.open()` creates a `pyvisa.ResourceManager` per driver and `close()` calls `rm.close()` on it (`instro/lib/transports/visa.py`). But pyvisa caches one `ResourceManager` per backend — `ResourceManager.__new__` returns the existing instance for the same VISA library. Closing it tears down the whole VISA session, invalidating every resource opened through it.

Consequences on any bench with 2+ VISA instruments on one backend (PSU + DMM is the canonical case):

- Closing one instrument kills every other instrument's session (`InvalidSession` on next I/O).
- `VisaDriver.__del__` calls `close()`, so a driver merely being garbage-collected can take down live instruments.
- `open()`'s failure path also closes the shared RM, so a failed *second* connect breaks the first instrument.

## Repro

Two `VisaDriver`s against `pyvisa-sim` (`@sim` backend), both answering `?IDN`. After `psu.close()`, `dmm.query()` raises `InvalidSession: Invalid session handle. The resource might be closed.` Unit tests don't catch this because each test patches `ResourceManager` per-test, so the singleton never appears.

## Fix

Stop closing the RM in `VisaDriver.close()` and in `open()`'s failure path; drop the `_rm` attribute (it existed only to be closed). pyvisa registers its own atexit handler that closes the manager — and any leftover sessions — at process exit. Each driver owns exactly one thing: its instrument session.

hoehner:
Fixes #78

## What

`VisaDriver` no longer closes the pyvisa `ResourceManager` — not in `close()`, and not in `open()`'s failure path. The `_rm` attribute is removed entirely; it existed only to be closed. Each driver now owns exactly one thing: its instrument session.

## Why

pyvisa caches one `ResourceManager` per backend (`ResourceManager.__new__` returns the existing instance for the same VISA library), so `rm.close()` tore down the whole VISA session shared by every instrument in the process:

- closing one instrument killed every other instrument's session (`InvalidSession` on next I/O)
- `__del__` calls `close()`, so garbage collection of one driver could take down live instruments
- a *failed* second connect closed the RM out from under the first instrument

pyvisa registers its own atexit handler that closes the manager (and any leftover sessions) at process exit, so nothing leaks.

## Verification

- Live repro with two `VisaDriver`s on the `pyvisa-sim` backend: before the fix, `dmm.query()` raised `InvalidSession` after `psu.close()`; after the fix the second instrument keeps answering.
- New regression test `test_close_does_not_break_other_driver_on_shared_resource_manager` — two drivers on one mock-shared manager; close the first, the second still reads. The existing fixture's `rm_class.return_value` naturally mirrors pyvisa's singleton, which is why per-test patching never caught this.
- Flipped the four assertions that pinned the old behavior (`rm_instance.close.assert_called_once()` → `assert_not_called()`) and renamed the two tests whose names described it.
- `just check` and `just test` pass (507 tests, mypy and ruff clean).


wharwood:
This looks good.

wharwood (approved):

PR #90, opened 2026-06-10
refactor: use instro.lib.publishers re-exports in examples and docs

hoehner:
Fixes INSTRO-348

The public publisher classes are already re-exported from `instro.lib.publishers`, but examples, tests, and docs still imported from the private submodules (`instro.lib.publishers.nominal_core`, `instro.lib.publishers.nominal_connect`). This switches all of those imports to the package re-exports.

- `examples/daq/*.py` (5 files) and `examples/eload/eload.py`: `NominalCorePublisher` now imported from `instro.lib.publishers`
- `tests/daq/mcc/test_mccdaq_hardware.py`: same
- `tests/lib/test_nominal_connect_publisher.py` and `docs/guides/instrumentation/publishers.mdx`: `NominalConnectPublisher` likewise (the `patch("instro.lib.publishers.nominal_connect.logger...")` targets are unchanged since patching must target the defining module)
- `docs/guides/instrumentation/examples/*.mdx` regenerated via `just gen-examples`
- `uv.lock` refreshed for the release-please version bumps from #57 (the lock was stale on `main`, so any `uv` invocation re-locked it)

`docs/reference/src/reference/publishers.md` is untouched: its `:::` directives document the modules themselves.

### DIFF
--- a/docs/guides/instrumentation/examples/daq/daq_analog_loopback.mdx
+++ b/docs/guides/instrumentation/examples/daq/daq_analog_loopback.mdx
@@ -9,7 +9,7 @@ import time
 
 from instro.daq import InstroDAQ
 from instro.daq.types import DAQVendor, Direction
-from instro.lib.publishers.nominal_core import NominalCorePublisher
+from instro.lib.publishers import NominalCorePublisher
 
 # Configuration: Choose your vendor.
 VENDOR = DAQVendor.LABJACK_T_SERIES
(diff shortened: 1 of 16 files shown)


dalvarez-204 (approved):
Looks good to me. Clean sweep of all import sites, patch() targets pointing at defining module. CI green across 3.10-3.13. GTM 🚀

PR #97, opened 2026-06-12
fix(daq): raise clear error when DAQ methods are called before open()

nhschwab:
## Summary

Calling an `InstroDAQ` device method before `open()` now raises a clear, typed `InstrumentNotOpenError` that names the instance, instead of surfacing a low-level transport/vendor-SDK error that doesn't point at the missing `open()` call.

## Changes

- **`instro/lib/exceptions.py`** — add `InstrumentNotOpenError(InstroError)`, mirroring the existing `FeatureNotSupportedError`; exported from `instro.lib`.
- **`instro/daq/daq.py`** — `InstroDAQ` tracks `_is_open` (set in `open()`, cleared in `close()`) and a `_require_open()` guard at the top of every device-touching method: `start`, `stop`, `read_analog`, `write_analog_value`, `write_digital_line`/`read_digital_line`, `write_digital_port`/`read_digital_port`, `close_relay`/`open_relay`, `get_points_in_buffer`. Message names the instance, e.g. `InstroDAQ 'myDAQ' is not open. Call open() first.`
- **`tests/daq/test_daq_drivers.py`** — parametrized not-open coverage across all guarded methods, plus message-names-instance, after-open, and close-re-arms tests. Existing functional tests now `open()` first to reflect the new contract.
- **`docs/guides/instrumentation/exceptions.mdx`** — document `InstrumentNotOpenError`. The reference doc auto-generates from the module docstring.

## Notes

- The guard lives at the `InstroDAQ` HAL (not the transport) because DAQ drivers use heterogeneous transports — VISA for SCPI, vendor SDKs for NI/LabJack/MCC — so a single transport-layer check wouldn't cover all of them.
- `configure_*` setup methods are intentionally left unguarded: the canonical flow configures after `open()`, and the ticket scopes the IO methods.
- INSTRO-76's declared blocker INSTRO-287 was canceled, but its outcome (drop `DAQDriverDataHandler`, drivers compose `VisaDriver`) has already landed — this change sits on that post-migration shape.

## Testing

- `just check` — ruff format, mypy, ruff lint all clean.
- `just test` — 834 passed.

### DIFF
--- a/docs/guides/instrumentation/exceptions.mdx
+++ b/docs/guides/instrumentation/exceptions.mdx
@@ -33,6 +33,30 @@ except FeatureNotSupportedError:
 
 Because it inherits from `InstroError`, `except InstroError` catches it.
 
+## InstrumentNotOpenError
+
+`InstrumentNotOpenError` (a subclass of `InstroError`) is raised when a device-touching method is called before `open()`. It points at the missing lifecycle call instead of surfacing a low-level transport or vendor-SDK error. The message names the instrument instance so you can tell which one you forgot to open.
+
+```python
+from instro.lib.exceptions import InstrumentNotOpenError
+
+daq = InstroDAQ(name="myDAQ", driver=...)
+try:
+    daq.read_analog()  # forgot to call daq.open() first
+except InstrumentNotOpenError:
+    ...
+```
+
+Open the instrument first — directly or via the context manager, which opens on entry and closes on exit:
+
+```python
+with daq:
+    daq.configure_analog_channel(...)
+    daq.read_analog()
+```
+
+Because it inherits from `InstroError`, `except InstroError` catches it.
+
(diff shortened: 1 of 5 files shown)


hoehner:
I think we should either work this ticket/PR globally (for all instruments at one time) or not at all. We can just straight up make the business rule that devices/instruments must be open() for us to talk to them in any way. The alternative (not doing this ticket) is also justifiable since its a little bit obvious that this is needed when considering our target audience for this library.

hoehner (changes requested):
all or nothin'

hoehner (approved):
After sidebar chats with Noah - I think we're fine to ship this as is. We won't prioritize applying this pattern across every other instrument for v1.0 but this use case does make sense at least for now.

PR #103, opened 2026-06-15
fix(lib): bundle every pyvisa-py backend for clean-install VISA support

### LINKED ISSUE #102: fix(lib): USB and GPIB instruments fail on a clean install under the @py backend
## Summary

On a clean `pip install instro`, USB and GPIB instruments cannot be opened. PR #92 switched the default pyvisa backend from `@ivi` to `@py` (`instro/lib/transports/visa.py:96`), but the only VISA dependency is bare `pyvisa-py` (`pyproject.toml:16`). pyvisa-py ships the *protocol* logic for USB and GPIB but not the backend libraries those buses need, so the sessions register as unavailable and `open()` fails.

## Impact

- **USB:** opening any `USB0::...::INSTR` resource raises a plain `ValueError: Please install PyUSB ...` (pyvisa-py needs `pyusb`). This is a regression from #92. Every shipped USB example breaks on a clean install: `examples/test_rack_example/test_rack.py`, all 8 `examples/daq/*.py`, `docs/guides/instrumentation/psu.mdx`, and the `VisaConfig` docstring's own example.
- **GPIB:** opening any `GPIB::...::INSTR` resource fails because pyvisa-py needs `gpib-ctypes` (and a native GPIB driver).
- The USB failure is a plain `ValueError`, not `VisaIOError`, so the troubleshooting table in `docs/guides/instrumentation/transports/visa.mdx` does not cover it. `docs/guides/instrumentation/driver-dependencies.mdx` also currently claims USB-TMC works without a vendor runtime, which is false as shipped.

## Fix

Pull the backend dependencies into the base install so both buses work (or fail clearly) out of the box:

- **USB — fully turnkey.** Use the `usb-full` extra. pyvisa-py's USB path falls back to `libusb_package`'s bundled libusb binary when no system libusb is present (`pyvisa_py/usb.py:34-54`), so `usb-full` (pyusb + libusb-package) makes USB work with zero native system setup.
- **GPIB — binding ships, native driver remains a prerequisite.** Use the `gpib-ctypes` extra so the binding installs and the code path imports cleanly. A native GPIB driver (NI-488.2, or linux-gpib + kernel module) is hardware-specific and cannot be bundled in a wheel; document the prerequisite and surface a clear message when it is missing.

Concretely: `pyvisa-py[usb-full,gpib-ctypes]>=0.8.1,<0.9` in `pyproject.toml`, plus docs corrections (`driver-dependencies.mdx`, `transports/visa.mdx`) and a regression test asserting the backend packages are installed.

## Out of scope (separate ticket)

`VisaDriver.clear()` raises `VI_ERROR_NSUP_OPER` on ASRL sessions under pyvisa-py 0.8.1 while its docstring prescribes it as the timeout-recovery step (`visa.py:232`). This is a serial-only issue unrelated to USB/GPIB support and will be filed separately to keep this PR focused.

Tracked in the pre-1.0 blocker review as ticket 42 (Tier 0 regression).

hoehner:
Closes #102.

## Problem

PR #92 switched the default pyvisa backend from `@ivi` to `@py`, but the only VISA dependency was bare `pyvisa-py`. pyvisa-py ships the protocol logic for its interfaces but none of the backend libraries the non-TCPIP ones need, so on a clean install opening any USB (`ValueError: Please install PyUSB ...`) or GPIB resource failed. Every shipped USB example broke.

## Fix

Depend on `pyvisa-py` with **every backend extra it offers**, so `@py` reaches NI-VISA-level interface coverage out of the box:

| Extra | Enables | Notes |
|---|---|---|
| `usb-full` | USB-TMC | bundles libusb via `libusb-package` — no system libusb needed |
| `gpib-ctypes` | GPIB | binding ships; a native GPIB driver (NI-488.2 / linux-gpib) is still a hardware prerequisite, now documented |
| `psutil` | TCPIP resource discovery | |
| `hislip-discovery` | HiSLIP mDNS discovery | |
| `vicp` | LeCroy VICP | |

`pyvisa-py[usb-full,gpib-ctypes,psutil,hislip-discovery,vicp]>=0.8.1,<0.9`

**Serial** is already covered by the existing `pyserial>=3.5` dependency (it also backs Modbus RTU), so it is not duplicated in the extras. **PXI/VXI** are the only interfaces a pure-Python backend cannot drive — those inherently still need a vendor runtime.

## Supported Python: 3.10–3.13

`requires-python` is tightened from `>=3.10,<4.0` to `>=3.10,<3.14` across the **top-level package and all seven workspace subpackages**, matching the CI test matrix and the supported set. Every backend package has wheels across that range, so the dependency is a single line with no environment markers.

(`requires-python` previously allowed 3.14+, where `libusb-package` has no wheel and this install fails — 3.14 never actually worked for USB, so scoping it out removes a false promise rather than real support. Revisit when `libusb-package` ships cp314 wheels.)

## Changes

- `pyproject.toml`: depend on all pyvisa-py backend extras + `requires-python` scoped to 3.10–3.13.
- `packages/*/pyproject.toml` (all 7): `requires-python` capped at `<3.14` to match the root. (`instro-ethernetip-rs` is a pure Rust crate with no `requires-python`.)
- `uv.lock`: regenerated (adds `pyusb`, `libusb-package`, `gpib-ctypes`, `psutil`, `zeroconf`, `pyvicp`, `ifaddr`). The subpackage caps do not change resolution — the workspace was already bounded by the root.
- `docs/.../driver-dependencies.mdx`, `transports/visa.mdx`: corrected to match what ships; GPIB native-driver prerequisite and PXI/VXI limitation documented.
- `tests/lib/test_visa_driver.py`: parametrized regression test asserting every backend package stays installed.

## Out of scope (separate ticket)

`VisaDriver.clear()` raises `VI_ERROR_NSUP_OPER` on ASRL sessions under pyvisa-py 0.8.1 (`visa.py:232`) — a serial-only issue unrelated to backend coverage. Filing separately to keep this PR focused.

## Verification

`just check` and `just test` pass on 3.13 (846 passed).


pyproject.toml
```diff
@@ -5,15 +5,20 @@ description = "Typed Python API for test-and-measurement instruments: power supp
 readme = "README.md"
 license = "Apache-2.0"
 license-files = ["LICENSE"]
-requires-python = ">=3.10,<4.0"
+requires-python = ">=3.10,<3.14"
```
nhschwab, line 8: We should cap the python version for the sub-packages and create an issue tracking the pending cp314 wheels for libusb

hoehner: Added and issue created to track 3.14 support https://github.com/nominal-io/instro/issues/105

nhschwab:
If we are capping at python version 3.13 we should make sure to document this well in the repo and on our docs page and explain why we have the cap

nhschwab (approved):

PR #110, opened 2026-06-16
fix(build): resolve Git Bash for just recipes on Windows

### LINKED ISSUE #109: just recipes fail on Windows: WSL bash shadows Git Bash
## Summary

On Windows, `just test` (and `just install`, `just eip-wheel-smoke-test`) fail with exit code 127 / 1:

```
/bin/bash: C:UsersjohnAppDataLocalTempjust-XXXXinstall: No such file or directory
error: Recipe `install` failed with exit code 127
```

## Root cause

These recipes are shebang recipes (`#!/usr/bin/env bash`). `just` runs them by writing the recipe body to a Windows temp file (`C:\Users\...\Temp\just-XXXX\install`) and invoking `bash <that path>`.

On a typical Windows dev machine `bash` resolves to **`C:\Windows\System32\bash.exe`** — the **WSL launcher** — which wins on `PATH` over Git Bash (`C:\Program Files\Git\bin\bash.exe`). WSL bash lives in a different filesystem namespace (it expects `/mnt/c/...`, not `C:\...`) and treats the backslashes as escape characters, so the path is mangled to `C:Usersjohn...` and the temp script can't be found.

Git Bash handles Windows-style paths correctly, so the recipes run fine once it takes precedence.

GitHub-hosted `windows-latest` runners also ship `C:\Windows\System32\bash.exe`, so this would block the planned Windows CI too.

## Fix

Prepend Git Bash to `PATH` in the `justfile`, conditional on OS, so plain `bash` resolves to Git Bash on Windows and is unchanged on Linux/macOS. Git Bash's directory is **derived from wherever `git` resolves on `PATH`**, rather than a hardcoded install path:

```just
export PATH := if os() == "windows" { (parent_directory(parent_directory(require("git.exe"))) / "bin") + ";" + env_var("PATH") } else { env_var("PATH") }
```

- `require("git.exe")` returns the full path to git (stable `just` builtin; aborts with a clear message if git is absent).
- Two `parent_directory` calls reach the Git install root — robust because Git for Windows ships `git.exe` in both `<root>\cmd` and `<root>\bin`, and two levels up from either lands on the root.
- `... / "bin"` is the Git Bash directory.
- `just` only evaluates the taken `if` branch, so `require("git.exe")` (which would fail on Linux/macOS, where git has no `.exe`) is never evaluated off-Windows; the `else` re-exports `PATH` unchanged.

### Why derived instead of hardcoded

An earlier draft hardcoded `C:/Program Files/Git/bin`, which breaks for anyone who installed Git to another drive or folder. Deriving from `git`'s actual location removes the drive/path assumption. The only remaining assumption is the standard Git-for-Windows directory layout (official installer, winget, Chocolatey); a bare scoop *shim* wouldn't resolve, which is a much narrower gap.

## Verification

`just test` passes end-to-end on Windows with the fix in place: 892 tests + the Rust suite (23 unit, 3 doctests) + the EtherNet/IP wheel smoke test, with no manual `PATH` tweaking.

## Notes

- Prerequisite for adding Windows + macOS CI.

hoehner:
## What

On Windows, `just test` / `just install` / `just eip-wheel-smoke-test` failed with exit 127:

```
/bin/bash: C:UsersjohnAppDataLocalTempjust-XXXXinstall: No such file or directory
```

These are shebang recipes (`#!/usr/bin/env bash`). `just` writes the body to a Windows temp file and runs `bash <path>`, but `bash` resolved to the **WSL launcher** (`C:\Windows\System32\bash.exe`), which expects `/mnt/c/...` and treats the backslashes as escapes — mangling the path so the script can't be found.

## Fix

Prepend Git Bash to `PATH` on Windows so plain `bash` resolves to Git Bash. The directory is **derived from where `git` resolves on `PATH`** (not a hardcoded install path), so it works regardless of which drive/folder Git is installed to:

```just
export PATH := if os() == "windows" { (parent_directory(parent_directory(require("git.exe"))) / "bin") + ";" + env_var("PATH") } else { env_var("PATH") }
```

`just` only evaluates the taken `if` branch, so `require("git.exe")` (which would fail on Linux/macOS) never runs off-Windows; the `else` re-exports `PATH` unchanged.

## Verification

`just test` passes end-to-end on Windows: 892 tests + Rust suite (23 unit, 3 doctests) + EtherNet/IP wheel smoke test, no manual `PATH` tweaking.

## Notes

- Prerequisite for adding Windows + macOS CI.
- Remaining assumption: the standard Git-for-Windows layout (official installer / winget / Chocolatey). A bare scoop shim wouldn't resolve.

Closes #109

### DIFF
--- a/justfile
+++ b/justfile
@@ -2,6 +2,10 @@ set positional-arguments := true
 
 export CARGO_TERM_COLOR := "always"
 
+# On Windows, ensure shebang recipes use Git Bash, not the WSL `bash` in System32 (see #109).
+# Git Bash's dir is derived from wherever `git` resolves on PATH, not a hardcoded install path.
+export PATH := if os() == "windows" { (parent_directory(parent_directory(require("git.exe"))) / "bin") + ";" + env_var("PATH") } else { env_var("PATH") }
+
 # Default command is no subcommand given to list available commands
 default:
     @just --list


joseph-gio (approved):

benliepert (approved):
Does `CONTRIBUTING.md` need to be updated to note that git bash is required to run `just` recipes on Windows?

PR #112, opened 2026-06-16
ci: add Windows and Mac to CI test matrix

### LINKED ISSUE #111: ci: add Windows testing to CI matrix
Extend the check-and-test workflow to run on Windows in addition to Linux, across Python 3.10-3.13.

- Add OS matrix (ubuntu-latest, windows-latest)
- Set bash as the default run shell so existing steps work on Windows runners
- Add .gitattributes (eol=lf) to avoid CRLF-induced ruff-format/cargo-fmt false failures
- Rely on runner-preinstalled Rust toolchain for the EtherNet/IP build

hoehner:
## What

Extends the `check-and-test` workflow to run on **Windows** alongside Linux, across Python 3.10–3.13 (8 jobs). macOS was intentionally skipped for now.

## Changes

* `.github/workflows/build-check-test.yml`
  * Matrix is now `os: [ubuntu-latest, windows-latest]` × `python-version: [3.10, 3.11, 3.12, 3.13]`, with `runs-on: ${{ matrix.os }}`.
  * Added job-level `defaults.run.shell: bash` so the existing bash-syntax steps (notably the `uv lock --check` guard) work on Windows runners, which default to PowerShell. GitHub ships Git Bash on Windows runners.
* `.gitattributes` (new)
  * `* text=auto eol=lf` so `ruff format --check` / `cargo fmt --check` don't flag CRLF on Windows. Renormalization touched zero existing files (repo was already LF), so there's no churn.

## Notes

* Rust is left to the runner's preinstalled toolchain (no `rust-toolchain.toml`). Accepted risk: if `cargo clippy -D warnings` fails on Windows only due to version drift, the follow-up fix is pinning the toolchain.
* `mcculw` (Windows-only) is not pulled in — `just install` uses no extras and `tests/daq/mcc` is ignored by default, so no cross-OS install issue.

Closes nominal-io/instro#111

### DIFF
--- a/.github/workflows/build-check-test.yml
+++ b/.github/workflows/build-check-test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13"]
     # Run every step under bash so the bash-syntax steps below (e.g. the
     # `uv lock --check` guard) work on Windows runners, which default to pwsh.
--- a/.github/workflows/build-check-test.yml
+++ b/.github/workflows/build-check-test.yml
@@ -11,11 +11,17 @@ on:
 
 jobs:
   check-and-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13"]
+    # Run every step under bash so the bash-syntax steps below (e.g. the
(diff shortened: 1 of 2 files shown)


hoehner:
Opted to also add macos to the mix here.

.github/workflows/build-check-test.yml
```diff
@@ -11,11 +11,17 @@ on:
 
 jobs:
   check-and-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
cartercanedy, line 18: depot runners?

cartercanedy (approved):

.github/workflows/build-check-test.yml
hoehner: Possibly eventually - for now just the standard runners that github offers.

PR #113, opened 2026-06-16
fix(daq): mcc fetch path integrity fix

nhschwab:
Fixes three defects in `MCCDriver.fetch_analog`'s continuous-scan path (`packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py`). Verified against the authoritative mcculw `get_status` docstring: `cur_count` is a signed 32-bit counter that advances in DMA **packet** increments (not multiples of `num_chans`), and `cur_index` — not `cur_count` — is the value documented to land on a completed channel-scan boundary.

## (a) Overrun realignment rotated channels permanently — blocker
On overrun the consume cursor was set to `curr_count - buffer_size`, which lands mid-scan because `cur_count` advances by packet size. That permanently offset `read_start`, so the per-sample gain mapping and the downstream de-interleave (`data[i::num_channels]`) reported every channel rotated — silently, behind only a printed warning. Now the realignment target is rounded **up** to the next multiple of `num_chans` (next full scan boundary; up, not down, so we never read into overwritten samples), keeping the cursor channel-aligned for all subsequent fetches.

## (b) Signed-32-bit count rollover caused permanent TimeoutError
In CONTINUOUS+BACKGROUND, `cur_count` rolls negative at 2^31 (~6 h at 100 kS/s aggregate), after which the cumulative-count comparison failed forever and every fetch raised `TimeoutError`. A new `_accumulate_count()` reinterprets the raw value as unsigned 32-bit (`& 0xFFFFFFFF`), detects the wrap, and accumulates a `2**32` offset to reconstruct a monotonic 64-bit count. State is reset in `start()`.

## (c) Torn copy near overrun returned mixed-generation samples
Overrun was only checked before the buffer copy, so DMA could overwrite the window mid-copy and return torn samples with no error. `fetch_analog` is now a retry loop: after the copy it re-reads `get_status` and discards/retries if the oldest valid sample advanced past the read origin. Consumption is committed only after a clean copy.

## Testing
All 14 hardware tests pass on a physical MCC USB-1616HS-4 (`tests/daq/mcc/test_mccdaq_hardware.py`), including the hw-timed fetch-path and analog-loopback cases that exercise the de-interleave/alignment path. `just check` and `just test` are green.

image

Closes INSTRO-384

### DIFF
--- a/packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
+++ b/packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
@@ -72,6 +72,8 @@ class MCCDriver(DAQDriverBase):
         self._buffer_size: int = 0  # Actual buffer size (fetch_size * multiplier)
 
         self._samples_consumed: int = 0  # Track consumed samples for streaming reads
+        self._raw_count_prev: int = 0  # Last raw (unsigned-32) cur_count, for rollover reconstruction
+        self._count_offset: int = 0  # Accumulated 2**32 rollovers of cur_count
         self._actual_sample_period: int = 0
         self._timestamper: HWTimestamper | None = None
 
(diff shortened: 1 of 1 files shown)


hoehner (approved):
Possibly worth adding a single unit test to lock in the behavior of the _accumulate_count function but not going to block on that for this PR merging. Everything else here looks solid.

PR #117, opened 2026-06-17
feat(daq): add Arduino Firmata DAQ driver

### LINKED ISSUE #49: driver: Arduino / Firmata (DAQ)
### Vendor

Firmata / Arduino

### Model

Arduino

### Category

DAQ (data acquisition)

### Protocol / transport

Raw serial

### Programming manual or datasheet

https://docs.arduino.cc/libraries/firmata/

### Can you contribute the driver?

Yes — someone on my team plans to.

### Additional context

General placeholder ticket to add Arduino devices under the "DAQ" category in instro. This ticket scopes adding the Firmata firmware on the Arduino while using one of the various python libraries to communicate to/from the device itself.

dalvarez-204:
Closes INSTRO-324. Fixes #49 

## Summary
- Adds ArduinoFirmata driver implementing DAQDriverBase via pyfirmata2 
- Supports analog input/pwm output, digital input/output
- Uses callback-based value streaming (pyfirmata2 does not support polling); daemon supported through callback-populated queue
- Hardware-timed acquisition not supported; 
- Sampling rate capped at 1000Hz (Firmata wire format limit)
- Adds 'arduino' optional extra to pyproject.toml

## Test plan
- [x] Manually validated on Mac, Linux, and Windows with real Arduino hardware

### DIFF
--- a/instro/daq/drivers/arduino_firmata.py
+++ b/instro/daq/drivers/arduino_firmata.py
@@ -34,7 +34,7 @@ class ArduinoFirmata(DAQDriverBase):
     Pin ranges depend on specific Arduino board.
     """
 
-    def __init__(self, port: str, sampling_rate_hz: float = 1000 / 19) -> None:
+    def __init__(self, port: str, sampling_rate_hz: float = 1000 / 19, buffer_size: int = 10_000) -> None:
         super().__init__()
         self._port = port
         if sampling_rate_hz > 1000:
@@ -49,10 +49,13 @@ class ArduinoFirmata(DAQDriverBase):
             raise ValueError(f"sampling_rate_hz must be > 0; got {sampling_rate_hz}Hz")
         else:
             self._sampling_interval_ms = int(1000 / sampling_rate_hz)
+
+        if buffer_size < 1:
+            raise ValueError(f"buffer_size must be >= 1; got {buffer_size}")
         self._board: pyfirmata2.Arduino | None = None
         self._pins: dict[str, Any] = {}
         self._latest_values: dict[str, float] = {}
-        self._sample_queue: queue.Queue[tuple[int, dict[str, float]]] = queue.Queue()
+        self._sample_queue: queue.Queue[tuple[int, dict[str, float]]] = queue.Queue(maxsize=buffer_size)
         self._pending_updates: set[str] = set()
         self._expected_ai_channels: frozenset[str] = frozenset()
 
@@ -89,6 +92,13 @@ class ArduinoFirmata(DAQDriverBase):
         self._sampling_interval_ms = int(1000 / rate_hz)
         if self._board is not None:
             self._board.setSamplingInterval(self._sampling_interval_ms)
+        if self._ai_hw_timing_config is not None:
+            sample_rate = 1000 / self._sampling_interval_ms
+            self._ai_hw_timing_config = HWTimingConfig(
+                sample_rate=sample_rate,
+                sample_period=round(1e9 / sample_rate),
+                samples_per_channel=1,
+            )
 
     def configure_ao_channel(self, channel: AnalogChannel) -> None:
         assert self._board is not None, "Call open() before configuring channels"
@@ -148,7 +158,16 @@ class ArduinoFirmata(DAQDriverBase):
         self._pending_updates.add(alias)
         if self._expected_ai_channels and self._pending_updates >= self._expected_ai_channels:
             timestamp = time.time_ns()
-            self._sample_queue.put((timestamp, dict(self._latest_values)))
+            if self._sample_queue.full():
+                warnings.warn(
+                    "ArduinoFirmata sample buffer is full; dropping oldest sample. "
+                    "Reduce sampling_rate_hz or increase buffer_size",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self._sample_queue.get_nowait()
+
+            self._sample_queue.put_nowait((timestamp, dict(self._latest_values)))
             self._pending_updates.clear()
 
     def read_analog(self) -> dict[str, float]:
--- a/instro/daq/drivers/arduino_firmata.py
+++ b/instro/daq/drivers/arduino_firmata.py
@@ -45,6 +45,8 @@ class ArduinoFirmata(DAQDriverBase):
             )
             self._sampling_interval_ms = 1
 
+        elif sampling_rate_hz <= 0:
+            raise ValueError(f"sampling_rate_hz must be > 0; got {sampling_rate_hz}Hz")
         else:
             self._sampling_interval_ms = int(1000 / sampling_rate_hz)
         self._board: pyfirmata2.Arduino | None = None
@@ -81,6 +83,9 @@ class ArduinoFirmata(DAQDriverBase):
                 stacklevel=2,
             )
             rate_hz = 1000
+
+        elif rate_hz <= 0:
+            raise ValueError(f"rate_hz must be > 0; got {rate_hz}Hz")
         self._sampling_interval_ms = int(1000 / rate_hz)
         if self._board is not None:
             self._board.setSamplingInterval(self._sampling_interval_ms)
@@ -151,7 +156,11 @@ class ArduinoFirmata(DAQDriverBase):
         return result
 
     def fetch_analog(self) -> dict[str, float]:
-        return self._sample_queue.get()
+        timeout = max(1.0, self._sampling_interval_ms / 1000 * 3)
+        try:
+            return self._sample_queue.get(timeout=timeout)
+        except queue.Empty:
+            return {}
 
     def _read_to_measurements(
         self,
@@ -161,6 +170,8 @@ class ArduinoFirmata(DAQDriverBase):
         default_tags: dict[str, str],
         **kwargs: Any,
     ) -> list[Measurement]:
+        if not response:
+            return []
         timestamp = time.time_ns()
         channel_data: dict[str, list[float]] = {}
         for alias, raw in response.items():
--- a/instro/daq/drivers/arduino_firmata.py
+++ b/instro/daq/drivers/arduino_firmata.py
@@ -22,18 +22,20 @@ if TYPE_CHECKING:
 class ArduinoFirmata(DAQDriverBase):
     """Arduino DAQ driver via Firmata protocol.
 
-    Arduino UNO:
-        Physical channel format - analog input: ''"A0"''-''"A5"'';
-        digital line: ''"D2"'' - ''"D13"''
+    Analog and digital I/O over StandardFirmata. The sampling rate controls how
+    often the Arduino sends analog readings back over serial. The Firmata default
+    is ~53 Hz (19ms interval). Rates above ~100 Hz may cause dropped messages at
+    the default 57600 baud; the actual ceiling depends on baud rate and the number
+    of configured channels.
 
+    Physical channel format: analog input "A0-A5"; digital line "D2"-"D13".
     Pin ranges depend on specific Arduino board.
-
     """
 
-    def __init__(self, port: str, sampling_interval_ms: int = 19) -> None:
+    def __init__(self, port: str, sampling_rate_hz: float = 1000 / 19) -> None:
         super().__init__()
         self._port = port
-        self._sampling_interval_ms = sampling_interval_ms
+        self._sampling_interval_ms = int(1000 / sampling_rate_hz)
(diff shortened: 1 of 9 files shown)


dalvarez-204:
image
image

Here are screenshots of windows and linux PCs demonstrating analog read, digital write, and digital read (this was also verified on mac)

dalvarez-204:
PWM is not implemented in this PR, but should be added later because it's a popular use case for arduinos. This DAQ implementation also relies on users knowing which analog and digital pins are available to them.

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,179 @@
...
+        self._board: pyfirmata2.Arduino | None = None
+        self._pins: dict[str, Any] = {}
+        self._latest_values: dict[str, float] = {}
+
+    def open(self) -> None:
+        try:
+            import pyfirmata2
+        except ImportError as e:
+            raise ImportError(
+                "pyfirmata2 is required for ArduinoFirmata.install it with: uv sync --extra arduino"
```
nhschwab:
```suggestion
                "pyfirmata2 is required for ArduinoFirmata. Install it with: uv sync --extra arduino"
```

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Arduino UNO:
+        Physical channel format - analog input: ''"A0"''-''"A5"'';
+        digital line: ''"D2"'' - ''"D13"''
+
+    Pin ranges depend on specific Arduino board.
+
+    """
+
+    def __init__(self, port: str, sampling_interval_ms: int = 19) -> None:
```
nhschwab: The sampling interval is currently only set here and can't be updated later. The only way to change the interval is to reconstruct the driver and `InstroDAQ` instance altogether. We can add a method to the driver to set the sampling interval which updates the internal private attribute. Not sure if you need to stop the active firmata sampling stream before updating the sampling interval, but something worth looking into.

Also why is the default sampling interval 19ms? And we should consider making the attribute sampling rate in Hz instead of sampling interval and doing an internal conversion. The firmata API takes interval in ms, but most engineers think abou sampling in rate Hz.

dalvarez-204: The 19ms is a recognized Firmata constant, and matches that set by the Arduino as a default. 

I understand the Instro library's whole point is that you should be able to just configure channels and call read_analog(), in which case Hz is the right unit because it's consistent with our DAQ ecosystem. My concern is that Arduino is the one device in this list where sampling rate is limited by the serial bandwidth rather than hardware, so stating it in Hz may give users a false sense of precision. For example, they might type 100 and wonder why their readings are getting dropped. 

In spirit of the instro lib, I'll go with Hz, but I'll make this constraint visible to the user. This way, users that desire abstraction have a familiar unit, and those who encounter streaming issues have a clear explanation of why

The interval can be updated without stopping the active firmata stream; I will implement this 👍

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,185 @@
...
+    def configure_ai_hw_timing(self, hw_timing_config: HWTimingConfig) -> None:
+        raise NotImplementedError("ArduinoFirmata does not support hardware-timed buffered acquisition")
+
+    def start(self, **kwargs: Any) -> None:
+        if self._board is not None:
+            self._board.samplingOn(self._sampling_interval_ms)
+
+    def stop(self, **kwargs: Any) -> None:
+        if self._board is not None:
+            self._board.samplingOff()
```
nhschwab, line 165: Start and stop are both explicitly for hardware-timed acquisition. The issue is that we still want to utilize firmata's streaming functionality and wire that directly into the background daemon so our publishers can continuously publish the streamed measurements without using a single point read + polling architecture. The current implementation in `InstroDAQ` requires the `ai_hw_timing_config` to be defined to successfully start the background daemon. The best solution I can think of would be:

1. Define a dummy `self._ai_hw_timing_config` at the driver level so that `daq.start()` does not raise
2. Instantiate a thread-safe, driver-mananged buffer 
3. Define `fetch_analog()` to drain the driver-manager buffer

This approach breaks with conventions and rules set by `InstroDAQ`, but it is the only way to cleanly implement the software-timed only firmata framework. `InstroDAQ` is meant to suppport DAQs with hardware-timed capability, so breaking some rules for an edge case software-timed only DAQ is fine. Just make sure it is heavily documented.

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def close(self) -> None:
+        if self._board is not None:
+            self._board.exit()
+            self._board = None
+        self._pins.clear()
+
+    def set_sampling_rate(self, rate_hz: float) -> None:
+        """Set the analog sampling rate. Rates above ~100Hz may cause dropped messages at the default 57600 baud."""
+        self._sampling_interval_ms = int(1000 / rate_hz)
```
nhschwab: make sure to protect against division by zero errors. Also any rate > 1000 will truncate to 0 with this solution.

dalvarez-204: My code does not handle it very gracefully, so I will fix that. Sub-millisecond sampling is not actually representable in the protocol. So the user cannot ask for quicker than 1ms (at least with StandardFirmata)

dalvarez-204:
Screenshot 2026-06-23 at 4 40 06 PM
Screenshot 2026-06-23 at 4 41 59 PM
image

Here are some of the results from the hardware testing. The first two images are from manual tests but the final image is from the arduino's hardware test file in the code.

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,240 @@
...
+        if sampling_rate_hz > 1000:
+            warnings.warn(
+                f"{sampling_rate_hz} Hz exceeds the Firmata protocol maximum of 1000Hz; clamping to 1000 Hz",
+                UserWarning,
+                stacklevel=2,
+            )
+            self._sampling_interval_ms = 1
+
+        else:
+            self._sampling_interval_ms = int(1000 / sampling_rate_hz)
```
nhschwab: Looks like the division by zero and negative rate is still unguarded?

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,240 @@
...
+            self._pending_updates.clear()
+
+    def read_analog(self) -> dict[str, float]:
+        result: dict[str, float] = {}
+        for alias in self._ai_channels:
+            result[alias] = self._latest_values.get(alias, 0.0)
+        return result
+
+    def fetch_analog(self) -> dict[str, float]:
+        return self._sample_queue.get()
```
nhschwab: Add a timeout and loop/return on an empty queue to avoid a hanging call in the background daemon

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,262 @@
...
+            self._sampling_interval_ms = 1
+
+        elif sampling_rate_hz <= 0:
+            raise ValueError(f"sampling_rate_hz must be > 0; got {sampling_rate_hz}Hz")
+        else:
+            self._sampling_interval_ms = int(1000 / sampling_rate_hz)
+        self._board: pyfirmata2.Arduino | None = None
+        self._pins: dict[str, Any] = {}
+        self._latest_values: dict[str, float] = {}
+        self._sample_queue: queue.Queue[tuple[int, dict[str, float]]] = queue.Queue()
```
nhschwab: This queue is unbounded which can lead to memory leak if the consumer thread slows down (e.g. core publisher slow due to network connectivity).  Add a max length to the queue and implement an evict-then-insert strategy in `_on_analog_callback` so that oldest samples in the queue are dropped and newest samples are retained when the queue is full. Add a warning statement whenever that happens so users know that samples are being dropped and provide a mitigation recommendation.

In practice this isn't a big deal since the arduino's max sample rate is relatively low, but it is a real risk we should mitigate. Choose a reasonable max number of samples for the queue, something like 10,000 (make sure you can justify the selection).

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,280 @@
...
+                warnings.warn(
+                    "ArduinoFirmata sample buffer is full; dropping oldest sample. "
+                    "Reduce sampling_rate_hz or increase buffer_size",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self._sample_queue.get_nowait()
+
+            self._sample_queue.put_nowait((timestamp, dict(self._latest_values)))
+            self._pending_updates.clear()
```
nhschwab, line 196: If the consumer (fetch_analog in daemon thread) accesses the queue in between the queue full check and the eviction, `self._sample_queue.get_nowait()` will thrown an error because the queue will be empty. We have to ensure that the daemon thread can drain the queue (even if 1 sample is lost) and the producer is tolerant to a newly empty queue. Something along the lines of:


```suggestion
    def _on_analog_callback(self, alias: str, value: float | None) -> None:
        self._latest_values[alias] = value if value is not None else 0.0
        self._pending_updates.add(alias)
        if self._expected_ai_channels and self._pending_updates >= self._expected_ai_channels:
            timestamp = time.time_ns()
            try:
                self._sample_queue.put_nowait((timestamp, dict(self._latest_values)))
            except queue.Full:
                warnings.warn(
                    "ArduinoFirmata sample buffer is full; dropping oldest sample. "
                    "Reduce sampling_rate_hz or increase buffer_size",
                    UserWarning,
                    stacklevel=2,
                 )
                with contextlib.suppress(queue.Empty):
                    self._sample_queue.get_nowait()   # evict; fine if consumer beat us to it
                with contextlib.suppress(queue.Full):
                    self._sample_queue.put_nowait((timestamp, dict(self._latest_values)))
            
            self._pending_updates.clear()
```

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def read_analog(self) -> list[tuple[int, dict[str, float]]]:
+        values = {alias: self._latest_values.get(alias, 0.0) for alias in self._ai_channels}
```
nhschwab: We shouldn't lie that a value is 0 if no data exists for that channel, we should raise instead

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,280 @@
...
+            bit_position=pin_num,
+        )
+
+    def write_digital_line(self, channel: DigitalChannel, data: int) -> None:
+        self._pins[channel.alias].write(data)
+        if self._board is not None:
+            self._board.sp.flush()
+
+    def read_digital_line(self, channel: DigitalChannel) -> int:
+        raw = self._latest_values.get(channel.alias, 0)
```
nhschwab: We shouldn't lie that a value is 0 if no data exists for that channel, we should raise instead

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,280 @@
...
+                f"{sampling_rate_hz} Hz exceeds the Firmata protocol maximum of 1000Hz; clamping to 1000 Hz",
+                UserWarning,
+                stacklevel=2,
+            )
+            self._sampling_interval_ms = 1
+
+        elif sampling_rate_hz <= 0:
+            raise ValueError(f"sampling_rate_hz must be > 0; got {sampling_rate_hz}Hz")
+        else:
+            self._sampling_interval_ms = int(1000 / sampling_rate_hz)
```
nhschwab:
```suggestion
            self._sampling_interval_ms =max(1, round(1000 / rate_hz))
```

instro/daq/drivers/arduino_firmata.py
```diff
@@ -0,0 +1,280 @@
...
+            warnings.warn(
+                f"{rate_hz} Hz exceeds the Firmata protocol maximum of 1000 Hz; clamping to 1000 Hz",
+                UserWarning,
+                stacklevel=2,
+            )
+            rate_hz = 1000
+
+        elif rate_hz <= 0:
+            raise ValueError(f"rate_hz must be > 0; got {rate_hz}Hz")
+        self._sampling_interval_ms = int(1000 / rate_hz)
```
nhschwab:
```suggestion
        self._sampling_interval_ms =max(1, round(1000 / rate_hz))
```

PR #128, opened 2026-06-18
fix(ethernetip): remove strings from Python EIP surface

benliepert:
context:
rust-ethernet-ip natively supports generic CIP STRINGs 
```
length : UINT   # 16-bit unsigned integer, 2 bytes
data   : octets # one byte per ASCII character
```
our CompactLogix L32E supports logix STRING tags
```
LEN  : DINT      # 32-bit signed integer, 4 bytes
DATA : SINT[82]  # 82 signed 8-bit integers, one byte each
```
## Summary
- References INSTRO-385: https://linear.app/nominal-io/issue/INSTRO-385/scope-out-strings
- Remove PLC string data types and string writes from the Python config-driven EIP API
- Remove string constructors/kinds/write acceptance from the private Python native EIP surface
- Update tests, type stubs, and docs to describe strings as out of scope for Python EIP
- Leave the Rust backend crate string support untouched

## Verification
- uv run pytest tests/test_ethernetip_config_data_type.py tests/test_ethernetip_config_timing.py tests/test_ethernetip_python_api.py -q
- uv run --no-cache --reinstall-package instro-ethernetip-python --with-editable . pytest tests/test_ethernetip_bindings.py -q
- just eip-wheel-smoke-test
- cargo clippy -p instro-ethernetip-python --all-targets --all-features -- -D warnings
- cargo test -p instro-ethernetip-python --all-features
- targeted ruff, mypy, and cargo fmt --check


docs/guides/instrumentation/protocols/ethernetip.mdx
```diff
@@ -183,48 +182,11 @@ The config-driven API supports scalar PLC tag types. The `data_type` field is re
 | `ulint` | `int` | Yes | 64-bit unsigned integer. Implemented, but not validated. |
 | `real` | `float` | Yes | 32-bit floating point. |
 | `lreal` | `float` | Yes | 64-bit floating point. |
```
cartercanedy, line 184:
```suggestion
| `lreal` | `float` | Yes | 64-bit floating point. |
| `string` | N/A | No | Allen-Bradley Logix `STRING` tags (`LEN` as `DINT`, `DATA` as `SINT[]`) are not supported |
```
or something like that

docs/guides/instrumentation/protocols/ethernetip.mdx
cartercanedy, line 1: I'd keep the strings section around as a reference, so we have documentation of our rationale for not supporting

packages/instro-ethernetip/instro/unstable/_ethernetip.pyi
cartercanedy, line 1: Should've asked this before, but are we generating these stubs or are these maintained by-hand? We should be leaning on automated tools so we don't wind up with incoherent interface stubs

cartercanedy (dismissed):

packages/instro-ethernetip/instro/unstable/_ethernetip.pyi
benliepert: they're manually generated rn
created [INSTRO-401](https://linear.app/nominal-io/issue/INSTRO-401/auto-generate-pyi-stubs)

cartercanedy (approved):

PR #131, opened 2026-06-22
feat(psu): config-driven InstroPSU initialization from JSON and dict

dalvarez-204:
## Summary

- Adds `PSUConfig` Pydantic v2 model (`instro/psu/config.py`) with validated fields for vendor, connection, num_channels, and visa_backend
- PSU can be constructed from json or dictionary 
- Exports `PSUConfig` from `instro.psu`
- Adds 6 unit tests, an example JSON config, and documentation

Closes INSTRO-75


instro/psu/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+class PSUConfig(BaseModel):
+    """Validated config for constructing an InstroPSU from JSON."""
+
+    model_config = ConfigDict(extra="allow")
+    name: str = Field(description="Channel-name prefix for published data.")
+    vendor: PSUVendor = Field(description="PSU vendor/model key.")
+    connection: str = Field(description="VISA resource string (e.g 'USB0::...' or 'TCPIP0::...').")
+    num_channels: int = Field(ge=1, description="Number of output channels.")
+    visa_backend: str = Field(default="@py", description="pyvisa backend specifier.")
```
nhschwab:
```suggestion
    visa_backend: str | None = Field(default=None, description="pyvisa backend specifier, defaults to @ivi and falls back to @py")
```
we want to maintain the `@ivi` default, `@py` fallback behavior

docs/guides/instrumentation/psu.mdx
```diff
@@ -55,6 +55,27 @@ psu = InstroPSU(
...
+```
+
+Where bench_psu.json contains:
+{
+  "name": "bench_psu",
+  "vendor": "bk_9115",
+  "connection": "USB0::0x2A8D::0x0101::MY12345::INSTR",
+  "num_channels": 1,
+  "dataset_rid": "<your dataset here>"
+}
```
nhschwab, line 112: wrap this with ```json

instro/psu/psu.py
nhschwab, line 1: We should add a `from_json_str()` classmethod so that a user can pass a json string directly. This will allow us to avoid deserializing json structs from the Connect UI

instro/psu/config.py
```diff
@@ -0,0 +1,77 @@
...
+    "tdk_lambda_genesys",
+]
+
+# now we make our little class for the PSU config
+
+
+class PSUConfig(BaseModel):
+    """Validated config for constructing an InstroPSU from JSON."""
+
+    model_config = ConfigDict(extra="allow")
```
nhschwab: extra should be should be "forbid" with explicit args to strictly enforce the model schema and error on bad args

instro/psu/config.py
nhschwab, line 1: Can we standardize the PSUVendor literal and the _registry dict to have one source of truth of supported drivers?

instro/psu/config.py
```diff
@@ -0,0 +1,77 @@
...
+    "bk_914x",
+    "keysight_e36100",
+    "keysight_n5700",
+    "rigol_dp800",
+    "siglent_spd3303",
+    "simulated",
+    "tdk_lambda_genesys",
+]
+
+# now we make our little class for the PSU config
```
nhschwab: delete

examples/psu/bench_psu.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "bench_psu",
+    "vendor": "bk_9115",
+    "connection": "USB0::0x2A8D::0x0101::MY12345::INSTR",
+    "num_channels":1,
+    "dataset_rid": "<dataset_rid"
+}
```
nhschwab: add a trailing line EOF

examples/psu/bench_psu.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "bench_psu",
+    "vendor": "bk_9115",
+    "connection": "USB0::0x2A8D::0x0101::MY12345::INSTR",
+    "num_channels":1,
+    "dataset_rid": "<dataset_rid"
```
nhschwab:
```suggestion
    "dataset_rid": "<dataset_rid>"
```

instro/psu/config.py
dalvarez-204: ok, this is done by making _registry: dict[PSUVendor, type], so now it's always checked against PSUVendor

instro/psu/config.py
```diff
@@ -0,0 +1,86 @@
...
+    model_config = ConfigDict(extra="forbid")
+    name: str = Field(description="Channel-name prefix for published data.")
+    vendor: PSUVendor = Field(description="PSU vendor/model key.")
+    connection: str = Field(description="VISA resource string (e.g 'USB0::...' or 'TCPIP0::...').")
+    num_channels: int = Field(ge=1, description="Number of output channels.")
+    visa_backend: str | None = Field(
+        default=None, description="pyvisa backend specifier, defaults to @ivi and falls back to @py."
+    )
+    dataset_rid: str | None = Field(default=None, description="Nominal dataset RID for auto-publishing.")
+    output_file: str | None = Field(default=None, description="File path for writing output data.")
```
nhschwab: This field is actually a directory, not a file. This could be confusing for users. Recommend renaming and updating field description

instro/psu/config.py
```diff
@@ -0,0 +1,86 @@
...
+        KeysightE36100,
+        KeysightN5700,
+        RigolDP800,
+        SiglentSPD3303,
+        SimulatedPSU,
+        TDKLambdaGenesys,
+    )
+    from instro.psu.psu import InstroPSU
+
+    _registry: dict[PSUVendor, type] = {
```
nhschwab: Tighten type to `type[PSUDriverBase]` to enforce PSUVendor -> PSUDriverBase mapping

nhschwab: We also need a way to enforce that new PSU drivers added in the library are tracked in the _registry and in the PSUVendor literal definition. Best way I can think of is to add a unit test that runs in CI but open to other ideas

dalvarez-204: unit test is a little tricky because we'd need _registry to be at the module level for the unit test to work, which would cause a circular import. I could work more with lazy imports but that solution's a bit messy. A half-solution would be to add an assertion inside build_psu_from_config itself. This wouldn't catch the mistake as early as a test, but it would run every time build_psu_from_config is called

nhschwab: We really want to avoid a test like this being at runtime, it should be part of CI as a unit test. We can achieve that by defining a module-level registry that maps strings to strings, where the keys are what you currently have defined as the PSUVendor literals and the values are string versions of the PSU driver classes. Then the vendor field in `PSUConfig` becomes a string type that we can validate with a field validator class method in `PSUConfig`. Then the second registry dictionary can go away and we can run a unit test without any circular imports.

stefanv:
Curious whether the goal is to make this more generic, so all instruments types can be configured from file?

examples/psu/bench_psu.json
hoehner, line 1: This config needs to look and feel as much like the modbus and ethernetip config as possible. 

That means starting with the following items at least:
```
{
  "version": 1,
  "protocol": "ethernetip",
  "device": {
    "name": "cell_controller",
    "description": "Sample ControlLogix/CompactLogix cell controller",
    "manufacturer": "Rockwell Automation",
    "model": "ControlLogix / CompactLogix"
  },
  "timing": {
    "poll_interval": 1.0
  },
...
```

hoehner: Actually this config is a better example in the docs
https://instro.nominal.io/instrumentation/protocols/modbus#config-my_device-json

hoehner:
> Curious whether the goal is to make this more generic, so all instruments types can be configured from file?

@stefanv - yes we're making an intentional push to have all instruments initialized from a config file - there are two other examples in the repo that do this (Modbus and EtherNet/IP). We ideally extend the same "config-driven workflow" to the other instruments we support in this library

nhschwab:
Claude is flagging that AGENTS.md needs to be updated to account for vendor-string factory repo constraints on line 111. We should either delete the constraint (this is a vestige from when we removed the auto discovery feature) or update the constraint so that it is clear the API does not include vendor-string factory but the config workflow is allowed to have it.

instro/psu/config.py
```diff
@@ -0,0 +1,130 @@
...
+    "SiglentSPD3303": "instro.psu.drivers.siglent_spd3303.SiglentSPD3303",
+    "SimulatedPSU": "instro.psu.drivers.simulated.SimulatedPSU",
+    "TDKLambdaGenesys": "instro.psu.drivers.tdk_lambda_genesys.TDKLambdaGenesys",
+}
+
+
+class TimingConfig(BaseModel):
+    """Timing configuration for PSU background polling."""
+
+    poll_interval: float = Field(ge=0.01, le=10.0, description="Polling interval in seconds")
```
nhschwab: Should we add `model_config = ConfigDict(extra="forbid")` here, too?

nhschwab: Also why is this capped on 10 seconds?

dalvarez-204: The polling interval cap mimicked what modbus had defined. Do you know of any documented upper bound we should utilize?

nhschwab: The InstroPSU API has no polling interval cap, so we should just follow that. Not sure about Modbus, might be protocol specific.

hoehner:
Couple things I would like to see:
1. The way to instantiate an InstroPSU from config should be InstroPSU(config="whatever.json"). This matches the convention created by Modbus and EthernetIP. I don't want to have to call an explicit .from_json() method. I know this is a python convention but I intentionally want it to be done in the constructor itself.
2. I need to see a PSU example where we launch a PSU from a config.
3. The config itself should have the additional parameters in the "device" section like the other drivers do. 
4. We should also have a way to fire up an InstroPSU from a PSUConfig object in code (like the other drivers).

hoehner (changes requested):
comment added!

instro/psu/psu.py
```diff
@@ -120,16 +123,20 @@ class InstroPSU(Instrument):
...
         self,
-        name: str,
-        driver: PSUDriverBase,
-        num_channels: int,
+        name: str | None = None,
+        driver: PSUDriverBase | None = None,
+        num_channels: int | None = None,
         publishers: list[Publisher] | None = None,
+        config: PSUConfig | dict | Path | str | None = None,
         **kwargs,
```
nhschwab, line 132: Can we also add the `autostart` argument like how it's done in modbus and eithernetip? Autostart meaning the instrument is opened and the background daemon polling thread is started.

docs/guides/instrumentation/examples/psu/psu.mdx
```diff
@@ -34,6 +34,8 @@ psu = InstroPSU(
 
 psu.add_publisher(NominalCorePublisher(dataset_rid=DATASET_RID))
 
+# Alternatively, construct from JSON config file (see bench_psu.json):
+# psu = InstroPSU.from_json("bench_psu.json")
```
hoehner: I think this comment is no longer valid syntax with the direct construction approach we've taken.

docs/guides/instrumentation/psu.mdx
```diff
@@ -59,13 +59,80 @@ psu = InstroPSU(
...
+- **`dataset_rid`**: Optional Nominal dataset RID keyword argument. When set, automatically publishes measurements to that dataset.
+
+### From a JSON Config File
+
+`InstroPSU` can also be constructed from a JSON config file, which removes the need to write any Python setup code: 
+
+```python 
+from instro.psu import InstroPSU
+
+psu = InstroPSU.from_json("bench_psu.json")
```
hoehner: Same here, we're going to always direct construct, not call from_json.

docs/guides/instrumentation/psu.mdx
```diff
@@ -59,13 +59,80 @@ psu = InstroPSU(
...
+```
+
+Where bench_psu.json contains:
+
+```json
+{
+  "version": 1,
+  "instrument": "InstroPSU",
+  "device": {
```
hoehner, line 89: This needs more fields to match the other drivers

docs/guides/instrumentation/psu.mdx
```diff
@@ -59,13 +59,80 @@ psu = InstroPSU(
...
+"publishers": [
+  { "type": "NominalCorePublisher", "dataset_rid": "<your dataset RID>" },
+  { "type": "FilePublisher", "directory": "psu_data", "format": "csv" }
+]
+```
+
+An optional top-level `timing` section (`{"poll_interval": 1.0}`) sets the background daemon's polling interval. To construct from a Python dict or a raw JSON string instead:
+
+```python
+InstroPSU.from_dict(data)
```
hoehner: Same here

docs/guides/instrumentation/psu.mdx
```diff
@@ -248,6 +315,9 @@ Every measurement/command call produces a channel keyed under `{name}.{descripto
 | `InstroPSU(name, driver, num_channels, publishers=None, legacy_naming=False, **kwargs)` | Construct an InstroPSU with a vendor driver |
```
hoehner, line 347: Update this to show our config argument

docs/guides/instrumentation/psu.mdx
```diff
@@ -248,6 +315,9 @@ Every measurement/command call produces a channel keyed under `{name}.{descripto
...
 | `open()` | Establish the connection to the PSU |
 | `close()` | Disconnect from PSU and close all publishers |
+| `from_json(path)` | Construct from a JSON config file |
```
hoehner: We shouldn't expose any of these

examples/psu/psu.py
```diff
@@ -29,6 +29,8 @@
 
 psu.add_publisher(NominalCorePublisher(dataset_rid=DATASET_RID))
 
+# Alternatively, construct from JSON config file (see bench_psu.json):
+# psu = InstroPSU.from_json("bench_psu.json")
```
hoehner: Update this to new format

instro/psu/psu.py
```diff
@@ -140,18 +147,87 @@ def __init__(
 
             num_channels: Number of output channels on this PSU.
             publishers: Publishers that receive emitted Measurement/Command data.
+                Combined with any publishers declared in ``config``.
+            config: A ``PSUConfig``, a dict, or a path to a JSON config file
+                (``str``/``Path`` is always a file path; for raw JSON text use
+                ``InstroPSU.from_json_str(json_str)`` instead).
```
hoehner: Remove this

hoehner (changes requested):

docs/guides/instrumentation/psu.mdx
```diff
@@ -59,13 +59,109 @@ psu = InstroPSU(
...
+        "name": "BK9115",
+        "num_channels": 1,
+        "connection_type": "visa",
+        "visa": {"visa_resource": "USB0::0x2A8D::0x0101::MY12345::INSTR"},
+    },
+    "timing": {"poll_interval": 1.0},
+    "publishers": [{"type": "NominalCorePublisher", "dataset_rid": "<dataset_rid>"}],
+})
+psu = InstroPSU(config=my_psu_config)
+```
```
nhschwab, line 165:
```suggestion
```python
# instantiated InstroPSU with plain dict config
psu = InstroPSU(config={
    "device": {
        "name": "bench_psu",
        "description": "Bench channel 1 supply",
        "manufacturer": "B&K Precision",
        "model": "9115",
    },
    "driver": {
        "name": "BK9115",
        "num_channels": 1,
        "connection_type": "visa",
        "visa": {"visa_resource": "USB0::0x2A8D::0x0101::MY12345::INSTR"},
    },
    "timing": {"poll_interval": 1.0},
    "publishers": [{"type": "NominalCorePublisher", "dataset_rid": "<dataset_rid>"}],
})

# instantiating InstroPSU with pre-built PSUConfig object
psu = InstroPSU(config=my_psu_config)
```
```

File: docs/guides/instrumentation/psu.mdx [hunk 20]
```diff
@@ -59,13 +59,109 @@ psu = InstroPSU(
...
+
+JSON configs are validated strictly: `PSUConfig` forbids any field not listed above (`device`, `driver`, `timing`, `publishers`), so there is no `**kwargs`-style escape hatch from JSON. Set default tags via the direct Python constructor instead.
```
nhschwab:
```suggestion
JSON configs are validated strictly: `PSUConfig` forbids any field not listed above (`version`, `instrument`, `device`, `driver`, `timing`, `publishers`), so there is no `**kwargs`-style escape hatch from JSON. Set default tags via the direct Python constructor instead.
```

docs/guides/instrumentation/examples/psu/psu_config.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: "three PSUs, three vendors, driven entirely by the JSON files next to this script"
```
nhschwab:
```suggestion
title: "PSU from JSON config"
```

docs/guides/instrumentation/psu.mdx
```diff
@@ -59,13 +59,109 @@ psu = InstroPSU(
...
+### Constructor Parameters
+
+- **`name`**: A name for this PSU instance. Used as a prefix for channel names when publishing. Falls back to `config.device.name` when `config` is given.
+- **`driver`**: A concrete `PSUDriverBase` instance (e.g. `SiglentSPD3303`) configured with the connection details for that model. Mutually exclusive with `config`.
+- **`num_channels`**: Number of output channels on the power supply. Mutually exclusive with `config`.
+- **`config`**: A `PSUConfig`, a dict, or a path to a JSON config file, as an alternative to `driver`/`num_channels`. See [From a JSON Config File](#from-a-json-config-file) below.
+- **`publishers`**: Optional list of publishers to attach. Combined with any publishers declared in `config`.
+- **`autostart`**: When `True`, opens the connection and starts background polling immediately.
 - **`**kwargs`**: Additional keyword arguments become default tags when using a publisher that supports tags (like `NominalCorePublisher`).
+- **`dataset_rid`**: Optional Nominal dataset RID keyword argument. When set, automatically publishes measurements to that dataset.
```
nhschwab: Pretty sure this is misleading and `dataset_rid` is actually included with kwargs?

examples/psu/psu_config.py
nhschwab, line 1: Can we make this example as simple as possible? The looping through config files and path resolution handling is distracting. The main point is showing how easy and simple it is to create an InstroPSU with a json.

docs/guides/instrumentation/psu.mdx
dalvarez-204: yes it is included with kwargs. I was thinking it might be good to differentiate it because its behavior is different from other kwargs: it triggers a publisher instead of becoming a tag. I agree its current position is somewhat misleading, so I'll make it a sub-item under kwargs

hoehner (approved):

PR #132, opened 2026-06-22
feat: add support for alicat mc flow controller and the concept of an InstroFlowController

smithednominal:
feat: add support for alicat mc flow controller and the concept of an InstroFlowController

## Summary

Adds a flow controller type to instro and a first driver for the alicat mc series.

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Serial communication with alicat MC-100SCCM-D-DB9M/5M at desk to test each function at the driver layer (alicat_mc.py) + executing example code for InstroFlowController

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers


smithednominal:
This commit https://github.com/nominal-io/instro/pull/132/commits/a44d7b7a1d1c747aa7dd8d88344810625c41c60b is a "big" change in terms of the interface. See commit message for logic. I think its the safe move but if you hate this concept, we can kill off that last trio of commits.

smithednominal:
now (artificially) blocked on https://github.com/nominal-io/instro/pull/173

packages/instro-unstable/instro/unstable/flowcontroller/flowcontroller.py
```diff
@@ -0,0 +1,195 @@
...
+        """Command a new flow setpoint in the device's configured engineering units."""
+        logger.debug("Sending FlowController set_setpoint to '%s'", self.name)
+        with self._resource_lock:
+            setpoint = self._driver.set_setpoint(value)
+            timestamp = time.time_ns()
+
+        return self._package_command("setpoint.cmd", setpoint, timestamp, **kwargs)
+
+    @publish_command
+    def select_gas(self, gas_name: str, **kwargs) -> Command:
```
smithednominal: rename to "set_media" or something

smithednominal:
Using this as a reference, further genericized the controller in a hopefully-not-painful way
https://documents.alicat.com/Alicat-Serial-Primer.pdf
basically each driver has to know what its process value actually is. For (to be implemented) liquid or pressure gauge controllers, this is constant. For the mass flow controller several options are permitted. In turn I added a function to set the process value (driver-level, not instro-level, as I assume most people will set by hand on the device UI).

This means the return type for get_flow_data has to basically just be a pretty generic dict because so many things can change at runtime, but I added as much type safety as I could (the dict must have setpoint + pressure, the only constants). There is more certainty at the driver layer because i made the mass flow controller return a mass-flow-controller-specific-type, but its still a dict under the hood. This aligns well with Measurement class but is a little weird. To make the weirdness easier to manage, I expose the most common values as properties including setpoint, pressure, and the generic "process value" (+ the key for what process value actually is). 

This will almost certainly hit a brick wall as soon as we implement another vendor, but for the time being I think its generic enough to work across the alicat line and probably works most of the way for another mass controller.

In addition to the alicat_mc, when I described the 3 different type variants it went ahead and scripted out a pressure controller and liquid controller. I did not (cannot) test them, but tbh I figured leaving them in was better than tossing them out, they just aren't bubbled up. Think of it as "contrib" but for "unstable" :) -- can delete if preferred.

smithednominal:
@hoehner i think this covered everything we discussed last week. See my most recent comment for the main updates (and files to potentially ignore or delete). going after the liquid controllers too was the move as it exposed the differences in devices while still being in the same style. I was thinking the "process value" query concept could also be reused if we wanted to wrap temperature controllers/watlows at some point too.

.vscode/settings.json
hoehner, line 1: Do we need to commit this file as a part of this PR?

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_liquid.py
```diff
@@ -0,0 +1,133 @@
...
+        """Serialize to the LiquidFlowData dict."""
+        return LiquidFlowData(
+            pressure=self.pressure,
+            temperature=self.temperature,
+            vol_flow=self.vol_flow,
+            setpoint=self.setpoint,
+        )
+
+
+class AlicatLiquidFlowController(FlowControllerDriverBase):
```
hoehner, line 46: I think this class is not instantiable because it doesn't implement "tare_flow" from the abstract base class. If the method has an @abc decorator then it must be overridden by the children classes. We can either make it optional (my preference) or implement it here. Also may be good to throw in a few tests to verify these can be instantiated!

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_pressure.py
```diff
@@ -0,0 +1,104 @@
...
+def _default_alicat_config(visa_resource: str) -> VisaConfig:
+    """Build default VisaConfig for Alicat RS-232 at 19200 baud."""
+    return VisaConfig(
+        visa_resource=visa_resource,
+        serial_config=SerialConfig(baud_rate=19200),
+        terminator=TerminatorConfig(read="\r", write="\r"),
+    )
+
+
+class AlicatPressureController(FlowControllerDriverBase):
```
hoehner, line 25: same as here https://github.com/nominal-io/instro/pull/132/changes#r3624846567

docs/guides/instrumentation/flowcontroller.mdx
```diff
@@ -0,0 +1,437 @@
...
+2. **`open()`**: establishes the VISA connection.
+3. **Configure and measure**: set voltage/current limits, enable outputs, read measurements.
+4. **`start()`**: begins a periodic background daemon. (Optional) By default this calls `get_flow_data` periodically.
+5. **`stop()`**: ends the background daemon (if started).
+6. **`close()`**: disconnects from hardware.
+
+## Creating an InstroFlowController Instance
+
+```python
+from instro.flowcontroller.drivers import AlicatMC
```
hoehner: Let's do a find and replace of all of these imports to include "unstable" 

instro.unstable.flowcontroller

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def select_working_fluid(self, fluid_name: str) -> str:
+        """Select the active working fluid (gas) by name and return the confirmed fluid name."""
+        return self._select_gas(fluid_name)
+
+    def _query_loop_variable(self) -> None:
+        """Query the device for its current loop control variable (process value source) and cache it."""
+        response = self._query_checked(f"{self.unit_id}LR{LOOP_VARIABLE_MASS_FLOW}")
+        fields = response.split()
+        # To query the current loop variable, use unit_idLR as the
```
hoehner, line 288: The comment here doesn't quite match the query sent above. Lets clean this up based on the real hardware behavior.

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
```diff
@@ -0,0 +1,527 @@
...
+            self._query_loop_variable()
+        sample = self._get_flow_data()
+        if self._cached_loop_variable == LOOP_VARIABLE_MASS_FLOW:
+            return sample.mass_flow
+        elif self._cached_loop_variable == LOOP_VARIABLE_VOL_FLOW:
+            return sample.vol_flow
+        elif self._cached_loop_variable in (LOOP_VARIABLE_ABS_PRESSURE, LOOP_VARIABLE_GAUGE_PRESSURE):
+            return sample.pressure
+        else:
+            return sample.mass_flow  # fallback to mass_flow if unknown
```
hoehner, line 401: Claude caught this one: 

_LOOP_VAR_TO_KEY includes PRESSURE_DIFFERENTIAL → "pressure_diff", but there's no branch for it here, so it falls through to mass_flow. On a differential-pressure device that means get_process_value() publishes a mass-flow value on a channel named <name>.pressure_diff

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
```diff
@@ -0,0 +1,527 @@
...
+        sum = GasMixEntry.sum_mixture_percentages(mixture)
+        if sum != 100:
+            raise ValueError(f"Gas mixture percentages to 2 decimal places must sum to 100, instead got {sum}")
+        mixture_strings = " ".join(
+            [f"{gas_entry.serialized_gas_percentage} {gas_entry.gas_number}" for gas_entry in mixture]
+        )
+        response = self._query_checked(f"{self.unit_id}gm {mix_name} {gas_id} {mixture_strings}")
+        response_cols = response.split()
+        mixture_identifier = GasTypeEntry(int(response_cols[1]), mix_name)
+        return mixture_identifier
```
hoehner, line 354: The gas cache goes stale here: after defining a mix, select_working_fluid("tstmix") raises "Unable to locate..." because known_gas_types still holds the pre-mix list. The user has to know to call list_gas_types(refresh=True) first.

Append the new GasTypeEntry to self.known_gas_types (or invalidate the cache) before returning.

docs/reference/src/instruments/index.md
```diff
@@ -10,6 +10,7 @@ instruments. Each instrument type defines a standard API that works across multi
 | [`InstroPSU`](psu.md) | Programmable power supply units |
 | [`InstroELoad`](eload.md) | Electronic loads |
 | [`InstroScope`](scope.md) | Oscilloscopes |
+| [`InstroFlowController`](flowcontroller.md) | Electronic loads |
```
hoehner: This looks like a typo here

docs/reference/src/instruments/flowcontroller.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Driver Interface
+
+::: instro.flowcontroller.FlowControllerDriverBase
+
+## Measurement keys
+
+`FlowControllerDriverBase` defines string constants for the keys returned by `get_flow_data()` and used by the single-value properties:
+
+| Constant | Key | Description |
```
hoehner, line 18: Can you verify these constant names are correct?

hoehner: I think they all end with _KEY now

docs/guides/instrumentation/flowcontroller.mdx
```diff
@@ -0,0 +1,437 @@
...
+## Key Concepts
+
+### Driver Composition
+
+An `InstroFlowController` is built from a concrete driver:
+
+```
+InstroFlowController(
+    name="FlowController",
+    driver = AlicatMC(VisaConfig(visa_resource=VISA_RESOURCE)
```
hoehner: Need an additional ")" here for this to work in real life

docs/guides/instrumentation/flowcontroller.mdx
```diff
@@ -0,0 +1,437 @@
...
+- The **vendor driver** (e.g. `AlicatMC`) owns the connection setup and vendor-specific command mapping.
+- **`InstroFlowController`** owns the category-level workflow: measurements, commands, publishers, the background daemon.
+
+### Lifecycle
+
+The typical InstroFlowController workflow:
+
+1. **Construct**: instantiate the vendor driver and pass it to `InstroFlowController`.
+2. **`open()`**: establishes the VISA connection.
+3. **Configure and measure**: set voltage/current limits, enable outputs, read measurements.
```
hoehner: Can we make this InstroFlowController specific here

packages/instro-unstable/instro/unstable/flowcontroller/examples/alicat_mc.py
```diff
@@ -0,0 +1,116 @@
...
+
+###Normal commanding
+setpoint = 1.23
+print(f"setting alicat setpoint to {setpoint}")
+setpoint_ret = device.set_setpoint(setpoint)
+print(f"Set alicat setpoint to {setpoint_ret}")
+print(f"Current flow reading = {device.get_flow_data()}")
+setpoint = 2.34
+print(f"setting alicat setpoint to {setpoint} using integer setpoint")
+setpoint_ret = device.set_setpoint_int(5.12, 100, 0)
```
hoehner: should this be setting the value assigned to "setpoint" ?

hoehner (changes requested):
Couple things from review!

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
```diff
@@ -0,0 +1,527 @@
...
+        self._visa.open()
+
+    def close(self) -> None:
+        """Close the VISA transport."""
+        self._visa.close()
+
+    def _query_checked(self, command: str) -> str:
+        """Query the device and raise RuntimeError if the response is '?'."""
+        response = self._visa.query(command)
+        if response == "?":
```
hoehner, line 204: Claude is telling me this is the default error that is returned for all errors. Lets ensure this propagates up to all callers so we don't stomp anything else that is wrapped up in this call. We want to make sure the OG error isn't lost.

docs/guides/instrumentation/flowcontroller.mdx
smithednominal: oof thats what I get for writing by hand. Do we have a CI step in (internal docs) for validating python snippets? For some reason I thought I saw that somewhere, might be worth bringing over here.

hoehner: I think we have a just test that validates the examples don't drift from the docs, but I don't think we have anything that verifies the syntax of the examples themselves :(

packages/instro-unstable/instro/unstable/flowcontroller/examples/alicat_mc.py
smithednominal: oh yep, i forgot to reset once I figured out the right denominator for this device (the int mechanism isn't well-documented :( )

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_liquid.py
smithednominal: Fair point -- I wasn't sure whether it would even be a good idea to leave these in since we can't test them with the hardware we have today. It feels weird to throw out, feels weird to include. I'll at least fix it up, though.

smithednominal: middle ground: added clear comment at the top of each file


WARNING: In contrast to the Alicat MC (mass flow controller), this is unstable, contrib-level code.
It has not been tested against actual hardware (the alicat MC has been tested and has a similar
but not identical interface). This driver is provided as-is and may require fixes when used with
a real Alicat liquid-flow controller.

(replacing with pressure for the other file of course)

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_pressure.py
smithednominal: resolved in same way

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
smithednominal: absolutely right. i dont know if you've seen this but sometimes claude caches something very wrong and even when you ask it to to something else it re-applies the wrong thing. A little frustrating, but I know the command I documented is correct. Will confirm on real hardware prior to merging, but I have manually confirmed this in the past.

smithednominal: fixed in bc15dbc3150cc141a0b4f6b4679c2b147fdf2c9e

smithednominal: for reference, page 15 here: https://documents.alicat.com/Alicat-Serial-Primer.pdf

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
smithednominal: Ah thats a fantastic catch. I think I'd prefer to re-query from the device so I'll invalidate the cache.

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
smithednominal: makes perfect sense in isolation, but actually not possible in practice (I think?). The class hierarchy is:

- FlowControllerDriverBase
  - Mass control <-where we are
  - Liquid
  - Pressure

Per the manual

> The controller can control the flow rate or the pressure in your process. Selectable control
loop variables include mass flow, volumetric flow, absolute pressure, and valve drive. Devices
with internal barometers also allow control of gauge pressure

So its arguable that falling through to the default is "right". However, I think its also fair to say thats a bit pedantic and the even more right answer is to handle this case, so I'll do so :)

smithednominal: 34851f6c71b6c40b27b1fe8ac7c17bc2d3f00d06

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
smithednominal: 34851f6c71b6c40b27b1fe8ac7c17bc2d3f00d06

packages/instro-unstable/instro/unstable/flowcontroller/drivers/alicat_mc.py
smithednominal: looks like this is correct but probably minor. The call chain where this fails is on tare, where I _expect_ a device may not support tare-ing and I make a _new_ exception without chaining to the existing. I don't think this will matter for most users (new error clearly says the feature isn't supported, which is right) but its definitely "wrong" and worth fixing.

smithednominal: a8e263c80d6ed6ee6a788497b7735f4bdf855f0e

.vscode/settings.json
smithednominal: removed

hoehner (dismissed):

hoehner (approved):

PR #136, opened 2026-06-22
fix(visa): catch pyvisa errors so @ivi→@py fallback fires

### LINKED ISSUE #133: fix(visa): @ivi→@py fallback never fires (VisaIOError is not OSError)
## Summary

The `@ivi` → `@py` backend fallback in `VisaDriver` (added in 61a2166, "feat(visa): default to @ivi backend, fall back to @py") never fires for the exact condition it was written to handle: a machine with no IVI/NI-VISA library installed. As a result, `VisaDriver.open()` crashes on any system without native VISA instead of falling back to pyvisa-py.

## Root cause

`_open_resource_manager` catches `OSError`, but a missing native VISA library surfaces as `pyvisa.errors.VisaIOError` (`VI_ERROR_LIBRARY_NFOUND`), whose MRO is `VisaIOError → pyvisa.errors.Error → Exception` — it is **not** an `OSError`. So the `except OSError` is skipped and the error propagates out of `open()`.

`instro/lib/transports/visa.py:278-291`:

```python
def _open_resource_manager(backend: str | None) -> pyvisa.ResourceManager:
    if backend is not None:
        return pyvisa.ResourceManager(backend)
    try:
        return pyvisa.ResourceManager(DEFAULT_VISA_BACKEND)
    except OSError as exc:   # <-- VisaIOError is not an OSError; never catches the missing-library case
        logger.warning(...)
        return pyvisa.ResourceManager(FALLBACK_VISA_BACKEND)
```

## Reproduction

On a machine with **no** NI-VISA / IVI installed but with `pyvisa-py` available, open any VISA driver with an unset backend (the default):

```
uv run python -c "import pyvisa; pyvisa.ResourceManager('@ivi')"
# pyvisa.errors.VisaIOError: VI_ERROR_LIBRARY_NFOUND (-1073807202): A code library required by VISA could not be located or loaded.
```

Confirmed against an Agilent 34401A over RS-232 (ASRL6) while writing its hardware-validation script — `InstroDMM.open()` crashed at RM construction before touching the serial port. Confirmed `pyvisa.ResourceManager("@py")` constructs cleanly and lists `ASRL6::INSTR` on the same machine.

## Suggested fix

Broaden the `except` to also catch the pyvisa error hierarchy:

```python
except (OSError, pyvisa.errors.VisaIOError) as exc:
```

(or `pyvisa.errors.Error` to be safe). Add a unit test that simulates the `@ivi` RM raising `VisaIOError` and asserts the fallback to `@py`.

## Workaround

Set `VisaConfig(visa_backend="@py")` explicitly to bypass the fallback path. The Agilent 34401A hardware-validation script does this for now.

hoehner:
## Summary

A missing native VISA library surfaces as `pyvisa.errors.VisaIOError` (`VI_ERROR_LIBRARY_NFOUND`), whose MRO is `VisaIOError → pyvisa.errors.Error → Exception` — it is **not** an `OSError`. So the `except OSError` in `_open_resource_manager` was skipped and the `@ivi`→`@py` fallback never fired on machines without IVI/NI-VISA installed; `VisaDriver.open()` crashed at ResourceManager construction.

This broadens the `except` to the pyvisa error hierarchy (`pyvisa.errors.Error`, the base of `VisaIOError`) alongside `OSError`.

## Test

Adds `test_open_falls_back_to_py_when_ivi_library_not_found`, which simulates the `@ivi` ResourceManager raising `VisaIOError(VI_ERROR_LIBRARY_NFOUND)` and asserts the driver falls back to `@py` and opens. Verified the test fails against the old `except OSError` and passes with the fix.

Fixes #133

### DIFF
--- a/instro/lib/transports/visa.py
+++ b/instro/lib/transports/visa.py
@@ -281,7 +281,7 @@ def _open_resource_manager(backend: str | None) -> pyvisa.ResourceManager:
         return pyvisa.ResourceManager(backend)
     try:
         return pyvisa.ResourceManager(DEFAULT_VISA_BACKEND)
-    except OSError as exc:
+    except (OSError, pyvisa.errors.Error) as exc:
         logger.warning(
             "VISA backend %s unavailable (%s); falling back to %s",
             DEFAULT_VISA_BACKEND,
--- a/tests/lib/test_visa_driver.py
+++ b/tests/lib/test_visa_driver.py
@@ -8,7 +8,8 @@ import time
 from unittest.mock import MagicMock, call, patch
 
 import pytest
-from pyvisa.constants import InterfaceType
+import pyvisa
+from pyvisa.constants import VI_ERROR_LIBRARY_NFOUND, InterfaceType
 from pyvisa.constants import Parity as VisaParity
 
 from instro.lib.transports import (
@@ -136,6 +137,18 @@ def test_open_uses_ivi_backend_by_default(mock_pyvisa):
 def test_open_falls_back_to_py_when_ivi_backend_missing(mock_pyvisa):
     rm_class, rm_instance, _ = mock_pyvisa
     rm_class.side_effect = [OSError("Could not locate a VISA implementation"), rm_instance]
+    driver = _make_driver()
+
+    driver.open()
+
+    assert rm_class.call_args_list == [call("@ivi"), call("@py")]
+    assert driver.is_open is True
+
+
+def test_open_falls_back_to_py_when_ivi_library_not_found(mock_pyvisa):
+    """A missing native VISA library surfaces as VisaIOError, not OSError (issue #133)."""
+    rm_class, rm_instance, _ = mock_pyvisa
+    rm_class.side_effect = [pyvisa.errors.VisaIOError(VI_ERROR_LIBRARY_NFOUND), rm_instance]
     driver = _make_driver()
 
     driver.open()


smithednominal (approved):
The change and tests makes sense (once I understood we were patching the underlying resource manager). No comments on that.

Suggestion for improvement:
I think it would be even _nicer_ if we took the fallback one step further to the actual resource itself. That is if a user specifies resource X it may be only accessible by one version of visa or the other. ASRL19::INSTR -> can only be used by NI Visa while the ASRLusbserial-AV0M67A2::INSTR format (I can't quite recall but its something like that on a Mac at least). So the sequence would become:

1. If user specifies a back-end => throw error if not present or if resource not available to backend
2. If user doesnt specify back-end: try resource name w/ ni visa then with pyvisa-py and use whichever of the two works first.

nhschwab:
> The change and tests makes sense (once I understood we were patching the underlying resource manager). No comments on that.
> 
> Suggestion for improvement: I think it would be even _nicer_ if we took the fallback one step further to the actual resource itself. That is if a user specifies resource X it may be only accessible by one version of visa or the other. ASRL19::INSTR -> can only be used by NI Visa while the ASRLusbserial-AV0M67A2::INSTR format (I can't quite recall but its something like that on a Mac at least). So the sequence would become:
> 
> 1. If user specifies a back-end => throw error if not present or if resource not available to backend
> 2. If user doesnt specify back-end: try resource name w/ ni visa then with pyvisa-py and use whichever of the two works first.

I agree with @smithednominal. Feels low priority but definitely an improvement. Mind making a ticket for this?

smithednominal:
https://github.com/nominal-io/instro/issues/138

PR #141, opened 2026-06-23
feat(visa): automatically switch between back-ends based on visa resource [closes #138]

### LINKED ISSUE #138: feat: Improve VISA backend fallback based on user-provided resource name
Resource fallback should take the specified resource itself into account for selecting a back-end if no explicit back-end is selected.

Some resource names are only accessible through one VISA implementation or the other. For example, `ASRL19::INSTR` may only work with `NI-VISA`, while a macOS USB serial resource such as `ASRLusbserial-AV0M67A2::INSTR` may only work with `pyvisa-py`.

Suggested behavior:

If the user specifies a backend and the resource doesn't exist for that backend, return an error if that backend is not available or if the resource is not accessible through that backend.

If the user does not specify a backend, try the provided resource name with `NI-VISA` first, then with `pyvisa-py`, and use whichever succeeds first. (or whatever the default priority order is)

This would make backend fallback more robust for resource formats that are backend-specific.

smithednominal:
closes #138 

## Summary

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

```python3
"""Diagnostic script: enumerate VISA devices via each backend, then verify open/fallback behavior.

Set ivi_index and pyvpy_index before running:
  ivi_index = index into default_resources (the @ivi list) to use for the @ivi-scenario tests.
  pyvpy_index = index into fallback_resources (the @py list) to use for the @py-scenario tests.
"""

import pyvisa

from instro.lib.transports import VisaConfig, VisaDriver
from instro.lib.transports.visa import DEFAULT_VISA_BACKEND, FALLBACK_VISA_BACKEND

# --- configure these before running ---
ivi_index = 2  # index into the @ivi resource list -- if you have a usb-to-serial dongle attached, 2 should work
pyvpy_index = 2  # index into the @py resource list -- usually = above


def _list(backend: str) -> tuple[str, ...]:
    try:
        rm = pyvisa.ResourceManager(backend)
        return rm.list_resources()
    except Exception as exc:
        print(f"  [enumerate {backend}] ERROR: {exc}")
        return ()


def _open(resource: str, backend: str | None) -> VisaDriver | None:
    label = backend if backend is not None else "(auto)"
    try:
        driver = VisaDriver(VisaConfig(visa_resource=resource, visa_backend=backend))
        driver.open()
        print(f"  [{label}] OPEN OK: {resource}")
        return driver
    except Exception as exc:
        print(f"  [{label}] OPEN FAILED: {exc}")
        return None


def _close(driver: VisaDriver | None) -> None:
    if driver is not None:
        try:
            driver.close()
        except Exception:
            pass


# ── 1. Enumerate ──────────────────────────────────────────────────────────────

print(f"\n=== Enumerate via DEFAULT backend ({DEFAULT_VISA_BACKEND}) ===")
default_resources = _list(DEFAULT_VISA_BACKEND)
for i, r in enumerate(default_resources):
    print(f"  [{i}] {r}")

print(f"\n=== Enumerate via FALLBACK backend ({FALLBACK_VISA_BACKEND}) ===")
fallback_resources = _list(FALLBACK_VISA_BACKEND)
for i, r in enumerate(fallback_resources):
    print(f"  [{i}] {r}")

# ── 2. Open index X from the @ivi list ────────────────────────────────────────

if not default_resources:
    print("\n[skip] No @ivi resources found; skipping open tests for X.")
else:
    target_ivi = default_resources[ivi_index]
    print(f"\n=== Open @ivi resource [{ivi_index}]: {target_ivi} ===")
    print("  Expected: OPEN OK / OPEN FAILED / OPEN OK")

    # Case 1: explicit @ivi — should work
    _close(_open(target_ivi, DEFAULT_VISA_BACKEND))

    # Case 2: explicit @py — should fail
    _close(_open(target_ivi, FALLBACK_VISA_BACKEND))

    # Case 3: no backend (auto-detect falls back correctly) — should work
    _close(_open(target_ivi, None))

# ── 3. Open index Y from the @py list ────────────────────────────────────────

if not fallback_resources:
    print("\n[skip] No @py resources found; skipping open tests for Y.")
else:
    target_py = fallback_resources[pyvpy_index]
    print(f"\n=== Open @py resource [{pyvpy_index}]: {target_py} ===")
    print("  Expected: OPEN FAILED / OPEN OK / OPEN OK")

    # Case 1: explicit @ivi — should fail
    _close(_open(target_py, DEFAULT_VISA_BACKEND))

    # Case 2: explicit @py — should work
    _close(_open(target_py, FALLBACK_VISA_BACKEND))

    # Case 3: no backend (auto falls through to @py) — should work
    _close(_open(target_py, None))

print()

```

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

Is there a good way to include that test script above? I don't think itll work w/ pytest unless there is hw attached on the runner.


smithednominal:
the verification script works still with 1 note: 

1. on windows, pyvisapy and ivi match names, so the things I expect to fail..don't
2. the script calls visa directly so the GPIB warning may still appear before the test cases run

instro/lib/transports/visa.py
```diff
@@ -309,6 +295,64 @@ def _open_resource_manager(backend: str | None) -> tuple[pyvisa.ResourceManager,
...
+        logger.info("Opening VISA resource %r via %s", visa_resource, DEFAULT_VISA_BACKEND)
+        try:
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", message=".*GPIB.*")
+                rm = pyvisa.ResourceManager(DEFAULT_VISA_BACKEND)
+            return typing.cast(
+                pyvisa.resources.MessageBasedResource,
+                rm.open_resource(visa_resource),
+            )
+        except (OSError, pyvisa.errors.Error) as exc:
```
nhschwab, line 328: this except catches everything pyvisa can raise when the default backend is tried, but the code assumes that the only reason the ivi backend could fail is because the device is not visible to ivi. Either gate on the errors that actually mean ivi can't see the device or capture the exception and included it in RuntimeError below so users know for certain why the open is failing

instro/lib/transports/visa.py
```diff
@@ -309,6 +295,64 @@ def _open_resource_manager(backend: str | None) -> tuple[pyvisa.ResourceManager,
...
+            ivi_resources = _list_resources(DEFAULT_VISA_BACKEND)
+            py_resources = _list_resources(FALLBACK_VISA_BACKEND)
+            raise RuntimeError(
+                f"VISA resource {visa_resource!r} not found via {DEFAULT_VISA_BACKEND} or {FALLBACK_VISA_BACKEND}.\n"
+                f"  {DEFAULT_VISA_BACKEND} resources: {ivi_resources}\n"
+                f"  {FALLBACK_VISA_BACKEND} resources: {py_resources}"
+            ) from exc
+
+
 def _is_missing_backend_error(exc: BaseException) -> bool:
```
nhschwab, line 361: This function is now dead code unless you reintroduce it to solve the comment above

instro/lib/transports/visa.py
```diff
@@ -309,6 +295,64 @@ def _open_resource_manager(backend: str | None) -> tuple[pyvisa.ResourceManager,
...
+def _open_resource(visa_resource: str, backend: str | None) -> pyvisa.resources.MessageBasedResource:
+    """Open a VISA resource with backend selection and fallback.
+
+    An explicit backend is used as-is. With no backend, tries ``@ivi`` first and
+    falls back to ``@py`` when the backend library is unavailable or the resource
+    string is not accessible through ``@ivi``.
+    """
+    if backend is not None:
+        logger.info("Opening VISA resource %r via %s", visa_resource, backend)
+        rm = pyvisa.ResourceManager(backend)
```
nhschwab: this now becomes susceptible to the gpib warning that we supressed in the previous change to the open resource logic, motivating the _open_resource_manager() function

nhschwab:
Left some comments that need attention, but overall this approach definitely improves UX. Make sure these changes are also reflected in CLI discover tool

instro/lib/transports/visa.py
smithednominal: sorry, you ended up reviewing code that I wasn't intending. I meant to use your function to catch the errors. Now fixed.

instro/lib/transports/visa.py
smithednominal: i think resolved now, error handling clearly delineates between failure mode

instro/lib/transports/visa.py
smithednominal: reintroduced

smithednominal:
> Left some comments that need attention, but overall this approach definitely improves UX. Make sure these changes are also reflected in CLI discover tool

unfortunately i had you review the wrong version of the code, sorry :(

one comment was cause of the wrong code, the other two were valid and I think i fixed.  

for the CLI it looks like its using your function to find the first available backend and then explicitly opening with that backend. that follows the unchecked exception path (below) so it should work as-is

```
if backend is not None:
        logger.info(f"Opening VISA resource {visa_resource!r} via {backend}")
        rm, _, _ = _open_resource_manager(backend)
        return typing.cast(
            pyvisa.resources.MessageBasedResource,
            rm.open_resource(visa_resource),
        )
```

PR #143, opened 2026-06-24
test(i2c): hardware validation for Total Phase Aardvark driver

hoehner:
## Summary

Verifies the Total Phase Aardvark I2C driver (`packages/instro-i2c-aardvark/instro/i2c/drivers/totalphase/aardvark.py`) against real hardware, per INSTRO-381.

Adds a standalone, `@pytest.mark.hardware` validation script at `tests/i2c/totalphase/test_aardvark_hardware.py` that exercises the driver across three phases against a Total Phase Aardvark adapter + [I2C/SPI Activity Board](https://www.totalphase.com/products/activity-board/):

- **Phase A — raw driver contract:** every method the driver implements — `open`, `close`, `set_bitrate`, `set_pullups`, `set_power_enable`, `read`, `write`, `write_read`.
- **Phase B — `SystemDefinition` + `I2CInterface`:** drives the same adapter through the public HAL with a real `SystemDefinition` (how users actually use the driver): register read/write roundtrips, field-level read-modify-write (which lights the board LEDs), and `reset_reg`.
- **Phase C — dense streaming:** runs the background daemon at a fixed rate (20 Hz), driving the PCA9554 output register with a sine and reading it back, so the test emits a dense (non-sparse) time series rather than a few scattered points.

Onboard targets: AT24C02 EEPROM @ `0x50` and PCA9554 I/O expander @ `0x38`. Enables target power + internal pull-ups before addressing the board and restores a safe state (target power off) in `finally`.

### Optional Nominal Core publishing

Mirrors the MCC hardware test's optional-publisher pattern: set **`NOMINAL_DATASET_RID`** to attach a `NominalCorePublisher` and stream the test's measurements/commands (including the Phase C sine) to a dataset — **no RID is hard-coded**, and credentials come from the on-disk `default` profile. When the env var is unset, no publisher is attached and the test still runs and validates fully. With it set, Core shows a clean commanded-vs-read-back sine confirming a faithful, dense I2C roundtrip.

## Tested instrument

Validation PASS (all phases) against a real unit:

| | |
|---|---|
| Serial / unique id | `2239-764425` |
| **Firmware version** | **3.51** |
| Hardware revision | 3.00 |
| API (SDK) version | 6.00 |

## CI note

The script is `@pytest.mark.hardware`, so it is **deselected by `just test` and in CI** (repo sets `addopts = "-m 'not hardware'"`). Run standalone:

```
uv run --extra i2c python tests/i2c/totalphase/test_aardvark_hardware.py
# optionally stream to Core:
NOMINAL_DATASET_RID=<rid> uv run --extra i2c python tests/i2c/totalphase/test_aardvark_hardware.py
```

`just check` (ruff format, mypy, ruff lint) passes. No driver or source changes — verification only.

Closes INSTRO-381.

Verification Evidence:
image


pyproject.toml
```diff
@@ -147,7 +147,10 @@ testpaths = ["tests"]
 # (mcculw is Windows-only) and must be invoked explicitly (e.g. `pytest tests/daq/mcc`).
 # --ignore=tests/daq/labjack: labjack tests depend on the optional `labjack` extra
 # (LJM SDK; hardware-only) and must be invoked explicitly (e.g. `pytest tests/daq/labjack`).
-addopts = "-m 'not hardware' --ignore=tests/daq/mcc --ignore=tests/daq/labjack"
+# --ignore=tests/i2c/totalphase: Aardvark tests depend on the optional `i2c` extra
+# (pyaardvark; hardware-only) and must be invoked explicitly (e.g.
+# `uv run --extra i2c python tests/i2c/totalphase/test_aardvark_hardware.py`).
+addopts = "-m 'not hardware' --ignore=tests/daq/mcc --ignore=tests/daq/labjack --ignore=tests/i2c/totalphase"
```
nhschwab, line 153: Non-blocking: this feels like it won't scale, should we consider a slight refactor to keep hardware tests organized and out of CI for now?

nhschwab (approved):

pyproject.toml
hoehner: I agree - I'll cut an issue to refactor how we have this setup in the repo.

hoehner: https://github.com/nominal-io/instro/issues/150

PR #159, opened 2026-06-25
feat(awg): adding AWGDriverBase and InstroAWG

### LINKED ISSUE #158: driver: adding driver for Rigol DG1022 (AWG)
**Vendor**
Rigol

**Model**
DG1022

**Category**
AWG (Arbitrary Waveform Generator)

**Protocol / transport**
SCPI over VISA

**Programming manual or datasheet**
https://www.batronix.com/pdf/Rigol/DG1022_DataSheet_EN.pdf

**Can you contribute the driver?**
Yes - I will be writing and testing the driver.

**Additional context**

The MVP scope for a new driver covers the standard control surface: lifecycle (`open`, `close`, `check_errors`), output configuration, and readback of key measurements. Extended features (protection settings, advanced modes, instrument-specific quirks) can be added in follow-on issues. A programming manual or datasheet is required to confirm SCPI tokens before implementation begins.

LizzyC-115:
# INSTRO-410: (Milestone 1) Add AWG driver base class and InstroAWG wrapper

Adds the abstract driver contract (`AWGDriverBase`), the Nominal publishing
wrapper (`InstroAWG`), waveform definition types, and a full software test
suite. No hardware driver is included — this establishes the interface that
concrete vendor drivers (e.g. Rigol DG1022Z) will implement in follow-on PRs.

## Files Added
- `packages/instro-unstable/instro/unstable/awg/awg.py`
- `packages/instro-unstable/instro/unstable/awg/types.py`
- `packages/instro-unstable/instro/unstable/awg/__init__.py`
- `tests/unstable/awg/test_awg.py`

## Design notes
- **Publishing**: `set_waveform` publishes the shape name as a string command
  (`ch1.waveform.cmd = "SQUARE"`) plus a companion all-float command carrying
  the numeric shape parameters (`ch1.frequency_hz.cmd`, …) at the same
  timestamp — numeric params stay plottable time series rather than string
  tags. `Arbitrary` samples are never published (only slot, count, and rate).
- **Docs**: deferred until the unstable API settles; will ship with the PR
  that promotes the category out of `instro-unstable`.

## Coverage — 108 tests, all passing

| Group | What's covered |
|---|---|
| Base class contract | abstract instantiation rules; optional-method `NotImplementedError` defaults |
| Waveform definitions | construction-time validation (frequency, duty cycle, symmetry, pulse width/delay vs period, sample normalization); immutability; `replace()` sweeps; empty-samples readback allowance |
| Lifecycle | `open`/`close` delegation; `start()` gating (0, 1-of-2, and all channels configured); daemon registration |
| Error checking | `check_errors` after every driver call; rejected `set_waveform` leaves no cached config; getter failures propagate |
| `set_waveform` | delegation; exact-type rejection (incl. subclasses); published-name vocabulary kept in sync with the `Waveform` union; companion numeric command (presence, values, shared timestamp, absence for `Noise`/`DC`); `Arbitrary` sample summarization and min-2-samples rule |
| Amplitude / offset / load / enable / phase | delegation, descriptors, unit-as-tag, float-only channel data, high-Z ↔ `float('inf')` mapping |
| Channel validation | bounds enforced on every channel-scoped method; 1 and `num_channels` valid, 0 and `num_channels+1` raise |
| Tag handling | user tags may reuse helper parameter names (positional-only helpers); tags pass through to published data |

## Notes
- No hardware required — all tests use `MagicMock(spec=AWGDriverBase)`.
- `get_waveform` and `get_amplitude` intentionally return raw values
  (`Waveform`, `tuple[float, AmplitudeMeasurementUnit]`) rather than published
  `Measurement`s — non-numeric return types.

## Validation
All 108 tests pass; `just check` clean (ruff format, mypy, ruff check).
`instro.unstable` is not yet in the root mypy `packages` list, so
`mypy -p instro.unstable.awg` was additionally run directly: clean.

Screenshot 2026-07-20 at 3 19 18 PM


packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,73 @@
...
+    PULSE = "PULSE"
+    NOISE = "NOISE"
+    DC = "DC"
+    ARB = "USER"
+
+
+# Start with 2 channels since Rigol DG1022 has 2 channels. More can be added if needed in the future.
+class Channel(Enum):
+    CH1 = 1
+    CH2 = 2
```
smithednominal: Consistency: I think most of the other drivers (at least the ones I have seen) use plain integers instead of an enumeration. I think its probably fine as long as you make sure the methods can just take a plain int as well.

Abstraction: The first time you make a thing you're definitely likely to make the abstract thing look like the real whether you want to or not, but explicitly limiting to 2 channels is probably too aggressive (if I were to leave this as an enum). I asked duck.ai to give me the biggest single-device arb it could find and it gave me 32 channels. (which also makes using an integer directly a bit more straightforward).

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,73 @@
...
+    ARB = "USER"
+
+
+class SweepSpacing(Enum):
+    LINEAR = "LIN"
+    LOGARITHMIC = "LOG"
+
+
+class TriggerSource(Enum):
+    INTERNAL = "IMM"  # DG1022 uses IMMediate, not INT
```
smithednominal: haven't gotten to the dg example code yet but I'd warn against leaking the abstraction here again. That having been said, I know scopes tend to call stuff "immediate" so its probably not wrong to do this.

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Makes sense! Updated to integers to allow for a broader scope

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,446 @@
...
+
+
+class AWGDriverBase(abc.ABC):
+    """Vendor AWG driver contract. Concrete drivers own their transport and lifecycle.
+
+    All methods here apply to standard waveforms. Composite waveform
+    support (modulation, sweep, burst, arb upload) is added in later milestones as
+    optional method groups that raise NotImplementedError by default.
+
+    ..._std_... - use with standard waveforms
```
smithednominal: it took me a minute to get this. I'd suggest:
1-Don't mention the future. It doesn't help anyone to have documentation on what might be :)
2-Make it clear you mean "methods with `_std_` in the name"...

Combined that gives you something like:

This driver supports configuration of standard pre-defined waveforms like sine, square, etc. These methods are delineated by having "...`_std_`..." in their name.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,446 @@
...
+    def open(self) -> None:
+        """Open the underlying transport."""
+
+    @abc.abstractmethod
+    def close(self) -> None:
+        """Close the underlying transport. Idempotent."""
+
+    @abc.abstractmethod
+    def check_errors(self) -> None:
+        """Query the instrument error queue and raise on non-zero error code."""
```
smithednominal: super minor, but don't call this a "non-zero error code" -- some instruments are dumb. We had to disable error handling in another driver because it returned "No Error" but in doing so caused an internal error...but in any case, no integer code, just a string.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,446 @@
...
+
+    # --- Standard periodic waveforms ---
+
+    @abc.abstractmethod
+    def set_std_waveform(self, channel: Channel, waveform: WaveformType) -> None:
```
smithednominal: I think individual methods here is good, I'm also wondering if theres a way to do a single atomic change w/in a driver? Like a transaction?

```
Begin Transaction
    set_wfm_type
    set_wfm_frequency
    set_wfm_amplitude
Commit
```

If not, is what it is...but you might in the higher level code* try to implement something like that where you (optionally) disable output, change several settings, re-enable output. For some applications, just changing a single param may be just fine (and desired, like to do a frequency sweep at a fixed amplitude). But in general I am anxious about the idea of having a function f(t,f,a) and not being able to change all three in a single atomic change.

*which you may have already done, I am a very linear reviewer

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,446 @@
...
+        """Set the waveform function on channel."""
+
+    @abc.abstractmethod
+    def get_std_waveform(self, channel: Channel) -> WaveformType:
+        """Get the current waveform function on channel."""
+
+    @abc.abstractmethod
+    def set_std_frequency(self, channel: Channel, frequency: float) -> None:
+        """Set the output frequency (Hz) on channel."""
```
smithednominal: Since you have a fixed set of `WaveformType`s, it might be good to indicate for which waveform types the frequency param applies. I *think* bolded ones below?

class WaveformType(Enum):
    **SINE = "SINE"**
    **SQUARE = "SQUARE"**
    RAMP = "RAMP"
    PULSE = "PULSE" **???**
    NOISE = "NOISE" **???**
    DC = "DC"
    ARB = "USER"

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,446 @@
...
+    @abc.abstractmethod
+    def set_phase(self, channel: Channel, phase_deg: float) -> None:
+        """Set the phase (degrees) for a channel."""
+
+    @abc.abstractmethod
+    def get_phase(self, channel: Channel) -> float:
+        """Get the current phase (degrees) for a channel."""
+
+    def align_phase(self) -> None:
+        """Sync the phase of both channels."""
```
smithednominal: I think its worth looking at other devices and seeing what options exist:

- do most devices have a "sync all channels" function
- or do most devices have a "sync channels X, Y"
If its the former, what you have is fine except replace "both" with "all"
if its the latter, I might make a param of list[int] for channels; if empty, assume "all" is desired.

smithednominal:
Lots of comments that I think are nice to haves. 

Since this is committing a new instro (without driver) I'd also add docs to _this_ pr. I will not claim my docs are great, but this may still serve as a pointer for what files need to change and where they live (anything .md, .json, .mdx): https://github.com/nominal-io/instro/pull/159

I think the key ones are:

- docs/guides/instrumentation/arbitrarywaveformgen.mdx (follow pattern)
- ../overview.mdx (add line for type)
- ../quickstart.mdx ~80 (add category to list in paragraph)
- docs/guides/index.mdx: add a <card> for your type
- docs/reference/src/instruments/arbitrarywaveformgen.md (follow pattern)
- ../index.md (one line addition for new type)
- docs/reference/src/index.md (one line for new type)
- docs/reference/mkdocs.yml (one line for new type)

tbh if you feed this comment to claude itll probably knock it out for you.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated to  """Vendor AWG driver contract. Concrete drivers own their transport and lifecycle.

    All methods here apply to standard waveforms and are delineated by __std__ in the method name. Solely composite/arbitrary waveforms are delineated by __arb__ in the method name.
    """

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: I added a `apply_waveform` function that a driver can implement to update a waveform and set all the parameters in one call if needed, in addition to the single parameter changes.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated the comments to show which waveforms each function is valid for. Will also update if hardware tests show other waveforms are possible inputs.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated to use just "error code"

tests/unstable/awg/test_awg.py
nhschwab, line 1: These tests should be part of CI - can we move to `tests/unstable/awg/` so that the CI config picks it up?

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,713 @@
...
+        Only applies when both channels are set to one of the following Enums:
+
+        - SINE
+        - SQUARE
+        - RAMP
+        - PULSE
+        - ARB
+
+        Per the programming manual, phase coupling is invalid when either channel is set to
+        NOISE or DC.
```
nhschwab: This is a confusing comment - this is supposed to be the abstract layer, not the device specific layer. Which programming manual are you referencing? I think this is universally true, so maybe delete this comment?

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,713 @@
...
+
+    def channel(self, n: int) -> int:
+        """Validate ``n`` is a valid channel number.
+
+        Raises:
+            ValueError: If ``n`` is outside the range 1–num_channels.
+        """
+        if n < 1 or n > self._num_channels:
+            raise ValueError(f"Channel {n} is out of range for '{self.name}' (1–{self._num_channels})")
+        return n
```
nhschwab: Is this method actually used anywhere? Would a user every actually use this method? Feels like this would only ever be used internally, in which case it should be a private method. But even then, it looks like this is unused anywhere and is dead code

packages/instro-unstable/CHANGELOG.md
nhschwab, line 1: changelog should not be manually edited, the release process handles updates

packages/instro-unstable/instro/unstable/awg/types.py
nhschwab, line 1: Double-tapping @smithednominal 's comment. It seems that a few of these enum fields are informed by the DG1022 SCPI command set, they should probably just remain generic

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,713 @@
...
+        - ARB (waveform_data is the uploaded sample array; per the programming manual,
+          this is the ``:TRACe``/``:DATA`` + ``:FUNCtion ARB`` upload-and-select path)
+
+        Args:
+            channel: Target output channel.
+            waveform: The waveform to apply.
+            waveform_data: The waveform data to apply.
+            disable_output: Whether to disable the output during waveform application.
+        """
+        raise NotImplementedError(f"apply_waveform is not implemented for {type(self).__name__}")
```
nhschwab: Looks like this driver method is not used anywhere in `InstroAWG`? Should we add a method that calls the driver method? Also docstrings say it only applies to ARB waveforms but the user can pass in any waveform type.

nhschwab:
Overall this is going in the right direction! One thing I would focus on:

What is the strategy for channel and waveform validation? Is it the responsibility of `InstroAWG` to validate that the right method is called on correctly configured channel? If so, you should add a framework to track the state of channel configurations. For example, if I call method on a channel that expects the channel to be configured with a certain waveform, how do we handle the case where the channel is an unexpected waveform? It would help to have a dict in `InstroAWG` or some defined struct that tracks the waveform configuration per channel.

packages/instro-unstable/CHANGELOG.md
LizzyC-115: Will revert to the CHANGELOG from main

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Got it, removed the comment that leaks the abstraction. Planning to use the enums as placeholders, then each driver can set their own specific codes in a dictionary or lookup table.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Removed to avoid dead code. Will add/reimplement in later milestones when it is relevant

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Removed

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated the comments to only describe the functionality of each method

cartercanedy:
@claude review

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,435 @@
...
+        raise NotImplementedError(f"set_ramp_symmetry is not implemented for {type(self).__name__}")
+
+    def set_pulse_width(self, channel: int, width_s: float) -> None:
+        """Set the pulse width (seconds) for a pulse waveform on channel."""
+        raise NotImplementedError(f"set_pulse_width is not implemented for {type(self).__name__}")
+
+
+_UNSET = object()
+
+_WAVEFORM_APPLICABILITY: dict[str, frozenset[WaveformType]] = {
```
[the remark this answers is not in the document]
LizzyC-115: Added a new validation method type to types.py.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,435 @@
...
+        return self._package_command(descriptor, value, timestamp, **kwargs)
+
+    @publish_command
+    def set_std_waveform(self, channel: int, waveform: WaveformType, **kwargs) -> Command:
+        """Set the waveform type on channel."""
+        self._check_channel(channel)
+        with self._resource_lock:
+            self._driver.set_std_waveform(channel=channel, waveform=waveform)
+            timestamp = time.time_ns()
+        self._channel_config[channel] = AWGChannelConfig(waveform)
```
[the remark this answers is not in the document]
LizzyC-115: Updated to obtain a lock before reading/writing to avoid a race condition.

packages/instro-unstable/instro/unstable/awg/awg.py
nhschwab, line 1: Looks like there is no `start()` method which pretty much all instrument support. Should be pretty straight forward and similar to `InstroDMM` with regards to how state is managed (`MeasurementFunction` vs. `AWGChannelConfig`

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,418 @@
...
+        if waveform in _WAVEFORM_APPLICABILITY[_ValidatedMethod.SET_STD_AMPLITUDE]:
+            cmds.append(self.set_std_amplitude(channel, amplitude, unit, **kwargs))
+        cmds.append(self.set_std_offset(channel, offset_v, **kwargs))
+        if load is not _UNSET:
+            cmds.append(self.set_std_output_load(channel, load, **kwargs))
+        if enable is not None:
+            cmds.append(self.output_enable(channel, enable, **kwargs))
+        if phase_deg is not None and waveform in _WAVEFORM_APPLICABILITY[_ValidatedMethod.SET_PHASE]:
+            cmds.append(self.set_phase(channel, phase_deg, **kwargs))
+        return cmds
```
nhschwab: I personally think we should remove this method. No other instrument implements a batch-configure method, and this is just wrapping existing methods in the instrument API. Better to let the user write their own code to wrap the commands than implement our own wrapper method that feels awkward

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,418 @@
...
+    def set_ramp_symmetry(self, channel: int, symmetry_pct: float) -> None:
+        """Set the symmetry (%) for a ramp waveform on channel."""
+        raise NotImplementedError(f"set_ramp_symmetry is not implemented for {type(self).__name__}")
+
+    def set_pulse_width(self, channel: int, width_s: float) -> None:
+        """Set the pulse width (seconds) for a pulse waveform on channel."""
+        raise NotImplementedError(f"set_pulse_width is not implemented for {type(self).__name__}")
+
+
+_UNSET = object()
```
nhschwab: This line is causing mypy fails due to typing. If we remove the configure_std_channel() method then this is no longer needed and can be removed, anyways

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def set_high_level(self, channel: int, volts: float, **kwargs) -> Command:
+        """Set the high voltage level (volts) on channel."""
+        self._check_channel(channel)
+        return self._execute_command(self._driver.set_high_level, channel, volts, "high_level", **kwargs)
+
+    def set_low_level(self, channel: int, volts: float, **kwargs) -> Command:
+        """Set the low voltage level (volts) on channel."""
+        self._check_channel(channel)
+        return self._execute_command(self._driver.set_low_level, channel, volts, "low_level", **kwargs)
```
nhschwab: Do these methods need to perform a waveform applicable check?

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,418 @@
...
+    SET_PHASE = auto()
+    SET_SQUARE_DUTY_CYCLE = auto()
+    SET_RAMP_SYMMETRY = auto()
+    SET_PULSE_WIDTH = auto()
+
+
+_WAVEFORM_APPLICABILITY: dict[_ValidatedMethod, frozenset[WaveformType]] = {
+    _ValidatedMethod.SET_STD_FREQUENCY: frozenset(WaveformType) - {WaveformType.NOISE, WaveformType.DC},
+    _ValidatedMethod.SET_STD_AMPLITUDE: frozenset(WaveformType) - {WaveformType.DC},
+    _ValidatedMethod.SET_PHASE: frozenset(WaveformType) - {WaveformType.NOISE, WaveformType.DC},
```
nhschwab: Phase is weird for pulse waveforms, some support and some do not, while others have a delay parameter. We should look into this and make sure that this level of abstraction works for all AWGs we may encounter

packages/instro-unstable/instro/unstable/awg/awg.py
nhschwab: Same thing with the background daemon definition

nhschwab: Also adding a form error checking from the instrument will be necessary, see other VISA instruments for how this is implemented

LizzyC-115: Updated to include a start() method that configures channels to begin with, a _define_background_daemon() method that will continuously read from the output state, and added a check_errors call to functions that require.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Agreed, removed the function call

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Added a variable that can be overwritten by the specific driver. Drivers with this capability can set the state to match what they allow.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: set_high_level and set_low_level shouldn't work for DC since they require amplitude, so added a waveform check to both functions. get_voltage_unit follows the other `get` functions in the file and does not check waveform.

LizzyC-115:
@claude review

LizzyC-115:
@claude review

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,29 @@
...
+from enum import Enum
+
+
+class WaveformType(Enum):
+    SINE = "SINE"
+    SQUARE = "SQUARE"
+    RAMP = "RAMP"
+    PULSE = "PULSE"
+    NOISE = "NOISE"
+    DC = "DC"
```
wharwood: Not thrilled to see DC here. DC voltages are not technically waveforms. This is one of those issues where this likely falls under the capability expressed in the PSU module. I think we need to figure out how to avoid grouping like capability in differing instruments. @hoehner, interested to get your feedback here.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,508 @@
...
+        """Close the underlying transport."""
+
+    @abc.abstractmethod
+    def check_errors(self) -> None:
+        """Drain the instrument error queue; raise if any error is pending."""
+
+    # --- Standard periodic waveforms ---
+
+    @abc.abstractmethod
+    def set_std_waveform(self, channel: int, waveform: WaveformType) -> None:
```
wharwood: This is going to apply to all the methods that have the _std_ marking. Where are we defining "std" waveforms? Is there a "non-std" method that we are planning to add to this? If so, why not push waveform definition into the WaveformType?

wharwood:
I'm seeing a whole lot of indexing on the features included in Function Generators. Given that this is an Arbitrary Waveform Generator interface, it should expose the ability to program _arbitrary_ waveforms. Personally, I'd prefer to keep the API much simpler. Most of the configuration is going to happen in the definition of the waveform itself. Once the user defines the waveform, the API for construction gets simple and has much more utility across the broad range of instruments: get/set waveform, amplitude, offset, output_enable, etc.

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@dataclass(frozen=True)
+class Sine:
+    """Sine waveform definition."""
+
+    frequency_hz: float
+    phase_deg: float = 0.0
+
+    def __post_init__(self) -> None:
```
wharwood, line 106: Any thoughts on truncating phase angle to +/- pi rad as part of post init validation?

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
...
+        """Validate shape parameters at definition time."""
+        _require_positive("frequency_hz", self.frequency_hz)
+        _require_percentage("duty_cycle_pct", self.duty_cycle_pct)
+
+
+@dataclass(frozen=True)
+class Ramp:
```
wharwood: I think "ramp" is much more commonly known as a Sawtooth wave.

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
...
+    """Ramp waveform definition."""
+
+    frequency_hz: float
+    symmetry_pct: float = 50.0
```
wharwood: Symmetry percentage is not a universal convention and will lead to confusion. You are trying to combine two waveforms with this parameter: sawtooth and triangle. I'd just remove this and implement a triangle wave. If a device combines them, the author of the implementation can handle calling the same shape and configure the correct parameters for that.

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
...
+    phase_deg: float = 0.0
+
+    def __post_init__(self) -> None:
+        """Validate shape parameters at definition time."""
+        _require_positive("frequency_hz", self.frequency_hz)
+        _require_percentage("symmetry_pct", self.symmetry_pct)
+
+
+@dataclass(frozen=True)
+class Pulse:
```
wharwood, line 161: Is a pulse not a square wave?

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.delay_s < 0:
+            raise ValueError(f"delay_s must be non-negative, got {self.delay_s}")
+        if self.width_s + self.delay_s >= 1.0 / self.frequency_hz:
+            raise ValueError(
+                f"width_s + delay_s must fit within the period (1/frequency_hz), got {self.width_s + self.delay_s}"
+            )
+
+
+@dataclass(frozen=True)
+class Noise:
```
wharwood: I think it's telling that Noise and DC don't have any parameters. I think there's a strong argument to be made to defer these until you have the interface definition more fleshed out. Noise is more important than a DC mode, IMO. But we need additional classes to define the types of noise. These vary depending on use case and have real meaning that we shouldn't expect the driver authors to implement.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -0,0 +1,316 @@
...
+        self._driver.close()
+        logger.info("Closed AWG '%s'", self.name)
+
+    @publish_command
+    def set_waveform(self, channel: int, waveform: Waveform, **kwargs) -> Command:
+        """Program channel with the waveform definition; numeric shape parameters publish as companion channels."""
+        if type(waveform) not in _PUBLISHED_NAMES:
+            raise TypeError(f"waveform must be a Waveform definition, got {type(waveform).__name__}")
+        self._check_channel(channel)
+        if isinstance(waveform, Arbitrary) and len(waveform.samples) < 2:
```
wharwood: The sample length check should probably move into post init validation of the waveform. This method allows the user to construct this and then have it blow up here when it is used.

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
+"""AWG shared types and waveform definitions."""
+
+from dataclasses import dataclass
+from enum import Enum
+
+
+class VoltageUnit(Enum):
```
wharwood: VoltageUnit is a poor name. Dbm is a unit of power. This should probably be renamed something like AmplitudeMeasurementType. It also misses Vp (half of Vpp for sine waves).

wharwood: One additional thought on these: the math to convert between these is universal. I'd be hesitant to allow drivers to implement a conversion system between them when we can expose that capability on the base class. On the other hand, if an instrument doesn't expose the capability to think in a certain amplitude metric, shouldn't we raise FeatureNotSupported? I'd tend to lean the former way myself, but this is open for discussion.

packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -0,0 +1,111 @@
...
+
+@dataclass(frozen=True)
+class Arbitrary:
+    """Arbitrary waveform definition; samples are normalized to [-1, 1] and scaled by amplitude/offset."""
+
+    # Requires >= 2 samples, enforced by InstroAWG.set_waveform.
+    samples: tuple[float, ...]
+    sample_rate_hz: float
+    # Use driver's default memory slot.
+    name: str = ""
```
wharwood: The name parameter seems unnecessary.

wharwood (changes requested):
Lots of minor things. The most pressing now is the decision regarding how to handle Noise and DC. Personally, I'd defer for this PR. There's nothing stopping us from addressing those at a later time while still providing the user base with good functionality now.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Moved to the post init

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Its a good idea. Simple implementation but keeps a consistent range across devices. Added a helper function to be called in post init

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Renamed to sawtooth

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: My thought was that square is a pulse with a set 50% duty cycle. We could fold the square into the pulse, but since the Rigol and the Keysight both have square and pulse separate, I added both.

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Agreed, it was a logic thing with having the sample check in awg, but resolved with moving that test to post init. Removed from this class

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Changed the name to AmplitudeMeasurementUnit

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Removed symm_pct and created a triangle class

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Using "StaticValue" instead of DC and removing Noise from the scope of this PR following the hardware-drivers talk

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Also, added conversion logic. We can always add a table that can cross check if a unit is readable to an instrument but the added logic would avoid reimplementing identical logic.

packages/instro-unstable/instro/unstable/awg/types.py
LizzyC-115: Agreed, removing noise from this PR scope and renamed DC to StaticValue with a value parameter.

packages/instro-unstable/instro/unstable/awg/types.py
wharwood: A square wave can have any duty cycle. The difference is the expression of the unit of that duty cycle. The pulse is being defined with an explicit timing whereas the square wave is defined as a percent of the duty cycle. I think we need to do a better job of documenting what the expectation is of our developers with respect to these. I can see a world where a pulse is expected to be a single period rather than a continuous one.

wharwood: More simply: these define the same shape of output but with different construction parameters.

wharwood (dismissed):
I think this is mostly covered. I just want a bit extra definition around what we expect the waveform types to define. See my in-line comments.

maxleblang (approved):
Approving to unblock due to docstrings change

PR #161, opened 2026-06-25
feat(opcua): implement browse paths

cartercanedy:
## Summary

This adds route-aware browse path handling, qualified browse names, cycle detection by ancestry, defensive browse limits, and updated client/harness coverage for browse behavior.

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [x] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

_*NOTE: will verify tests once integration is done*_

## Tests

- [ ] No tests — explain why:

## Checklist


cartercanedy:
@Leundai any idea what's going w/ policy-bot here?

packages/instro-opcua/src/browse.rs
```diff
@@ -128,21 +174,49 @@ fn browse_recursive<'a, B: Browse>(
...
+                bail!(
+                    "cycle detected while browsing node {} at path {}",
+                    node.node_id,
+                    parent_path
+                );
             }
 
-            if matches!(node.node_class, OpcUaNodeClass::Object) {
+            *visited = visited.saturating_add(1);
+            if *visited > max_nodes {
```
benliepert: nit: this check will never be hit if `max_nodes == usize::MAX` because the saturating addition will wrap around. should move the check prior to the saturating add and use `>=`

crates/instro-opcua/src/types.rs
```diff
@@ -564,12 +564,225 @@ pub enum NodeIdInner {
     String(String),
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
benliepert, line 568:
```suggestion
/// One segment in a browse path, preserving the namespace that qualifies the name.
///
/// OPC UA browse names are only unique within a namespace, so two nodes may both
/// be named `Pressure` while referring to different qualified names such as
/// `2:Pressure` and `4:Pressure`.
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

crates/instro-opcua/src/types.rs
```diff
@@ -564,12 +564,225 @@ pub enum NodeIdInner {
...
+impl QualifiedBrowseName {
+    pub fn new(namespace_index: u16, name: String) -> Self {
+        Self {
+            namespace_index,
+            name,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
benliepert, line 587:
```suggestion
/// Route to a browsed OPC UA node, represented as namespace-qualified browse-name segments.
///
/// Each segment preserves its namespace index so paths remain distinguishable when
/// different namespaces contain the same browse name.
#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

benliepert (changes requested):

crates/instro-opcua/src/client.rs
```diff
@@ -445,16 +537,11 @@ impl OpcUaClient {
...
-
             if let Some(samples) = samples {
                 on_data(Box::new(samples.into_iter()));
             }
 
-            match polling_interval.checked_sub(start_time.elapsed()) {
-                Some(duration) => sleep(duration).await,
-                None => yield_now().await, // injected suspension point to allow cooperative cancellation before next server read
-            }
+            metrics.finish_read(read_start, outcome);
```
benliepert: why'd you move this after? now it includes the time it takes to run the `on_data` callback

crates/instro-opcua/src/types.rs
```diff
@@ -564,12 +564,227 @@ pub enum NodeIdInner {
...
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct OpcUaNode {
     pub node_id: OpcUaNodeId,
     pub browse_name: String,
     pub display_name: String,
     pub node_class: OpcUaNodeClass,
+    #[serde(default)]
+    pub browse_path: BrowsePath,
```
benliepert, line 795: This is a breaking change since the struct and members are `pub`
user code could have e.g.
```rust
let n = OpcUaNode {
    node_id,
    browse_name,
    display_name,
    node_class,
    children,
}
```
which would fail to compile since `browse_path` is now required

cartercanedy: since there's a bunch of other rather large changes, not super worried about breaking api. Gonna land them all and stabilize in a major version bump

crates/instro-opcua/src/client.rs
cartercanedy: artifact of an experiment, reverting

packages/instro-ethernetip/Cargo.lock
benliepert, line 1: revert? unrelated to PR afaict

benliepert (dismissed):
add docs for new types

packages/instro-ethernetip/Cargo.lock
cartercanedy: see
- https://github.com/nominal-io/instro/issues/267
- INSTRO-469

cartercanedy: you're actually super right about most of that, codex went off the rails on the merge resolution. lockfile update is coming in another PR

hoehner (approved):

PR #172, opened 2026-06-26
test(daq): validate MCC USB-1616HS-4 driver against hardware

nhschwab:
## INSTRO-380 — Verify MCC DAQ

Verifies the `MCCDriver` DAQ driver (`packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py`) against a real **MCC USB-1616HS-4** and lands the standalone hardware-validation script used to do it.

### Tested instrument
- **Model:** MCC USB-1616HS-4 (serial redacted), product_id=205
- **Firmware version: N/A** — the USB-1616HS-4 exposes no version through any mcculw path: `get_config_string(BOARDINFO, DEVVERSION, <every FirmwareVersionType>)` returns `''` and `get_config(BOARDINFO, FWVERSION, ...)` returns `0x0000`. Recorded as N/A; a device/UL reporting gap, not a driver issue.

### Loopback wiring used
Analog (differential AI):
- `VOUT0 -> CH0H/CH0L` (AI ch 0)
- `VOUT1 -> CH1H/CH1L` (AI ch 1)

Digital (only bits 0-1 wired):
- `FIRSTPORTA bit 0 -> FIRSTPORTB bit 0`
- `FIRSTPORTA bit 1 -> FIRSTPORTB bit 1`

### Coverage — 21 checks, all passed
| Group | Methods |
|---|---|
| Lifecycle / device info | `open`, `close`, `get_info`, firmware / serial read |
| Software-timed AI | `configure_analog_channel(INPUT)`, `read_analog` — 2 differential channels, finite assert |
| Analog out | `configure_analog_channel(OUTPUT)`, `write_analog_value` — ao_0 + ao_1 voltage sequence |
| Analog loopback (SW) | ao_0->ai_0 and ao_1->ai_1, 6-point sweep, both channels within 0.15 V |
| Narrow range (SW) | ai_0 at +/-1 V range, 0.5 V readback within 0.15 V |
| Boundary voltages (SW) | ai_0 across -9.5...+9.5 V, within 0.15 V |
| HW-timed (background) | `configure_ai_sample_rate`, `start()`, `get_channel` — mean ai_0/ai_1 at +/-3 V, 0.15 V tol |
| HW-timed (foreground) | `start(background=False)`, `read_analog` |
| HW-timed ramp loopback | ramp on ao_0/ao_1, means converge to endpoint, 0.15 V tol |
| Sample-rate telemetry | `get_actual_sample_rate` within 10% of 1 kHz |
| Sustained acquisition | 3 s at constant 2 V DC, std dev < 0.1 V (overrun / stability check) |
| Buffer-depth telemetry | `get_points_in_buffer` non-negative finite during background acquisition |
| Daemon buffer guard | `read_analog` raises `RuntimeError` while the background daemon is active |
| Command return value | `write_analog_value` -> `Command` channel key + value assert |
| Unconfigured channel guard | `write_analog_value` on unregistered alias raises `KeyError` |
| Digital port config | `configure_digital_port` — FIRSTPORTA (OUT) + FIRSTPORTB (IN), 8-bit |
| Digital port loopback | `write_digital_port`/`read_digital_port` — 2-bit patterns 00/01/10/11 |
| Digital port full byte | 0x00/0xFF/0xAA/0x55, looped-back bits 0-1 verified |
| Clean shutdown | analog outputs -> 0 V, port -> 0x00 via public API |
| Per-line digital config | `configure_digital_line` raises `RuntimeError` (USB-1616HS-4 has no `d_config_bit`) |
| Relay control | `configure_relay_channel` + `close_relay` raises `NotImplementedError` |

### Notes
- **No driver changes were needed** — the driver behaved correctly on every implemented method.
- The USB-1616HS-4 does not support per-line digital configuration (`d_config_bit`), so digital I/O is exercised through the port-width API; the line path is asserted to raise.
- `read_analog` during background acquisition raises `RuntimeError` by design (buffer ownership, INSTRO-149); test 16 asserts this.
- The firmware read now uses the correct mcculw API (`get_config_string` + `BoardInfo.DEVVERSION` + `FirmwareVersionType`); this unit simply reports nothing.
- Device serial and `DATASET_RID` are redacted; `DATASET_RID` defaults to `None` (publishes nowhere) and `DEVICE_ID` must be set to the unit serial to run.
- The script is `@pytest.mark.hardware`, so it is **deselected in CI** (`-m 'not hardware'`) and only runs against a connected unit.
- Run standalone with `uv run pytest tests/daq/mcc/test_mccdaq_hardware.py -m hardware -v -s`.

### Checks
- `ruff format` + `ruff check` — clean
- `mypy` — clean
- `uv run pytest` — 993 passed, 271 deselected, 1 xfailed (the 21 hardware checks deselected)
- Hardware suite — **21/21 passed** on the connected USB-1616HS-4

### Verification

image

image

### DIFF
--- a/tests/daq/mcc/test_mccdaq_hardware.py
+++ b/tests/daq/mcc/test_mccdaq_hardware.py
@@ -42,7 +42,7 @@ NOMINAL CORE CONFIGURATION
 
   Before running, set the following environment variables:
 
-    MCC_DEVICE_ID       — unique device ID reported by mcculw (e.g. "344371")
+    MCC_DEVICE_ID       — unique device ID reported by mcculw / InstaCal (the device serial)
     NOMINAL_DATASET_RID — dataset RID for the NominalCorePublisher
     NOMINAL_API_TOKEN   — Nominal API token (optional if authenticated via
                           `nominal auth set-token`, which stores a default profile
@@ -66,6 +66,8 @@ import unittest
 from datetime import timedelta
 
 import pytest
+from mcculw import ul
+from mcculw.enums import BoardInfo, FirmwareVersionType, InfoType
 from nominal.core import EventType, NominalClient
 
 from instro.daq import InstroDAQ
@@ -76,8 +78,13 @@ from instro.lib.publishers import NominalCorePublisher
 # ---------------------------------------------------------------------------
 # Configuration from environment
 # ---------------------------------------------------------------------------
-DEVICE_ID = "344371:0"  # MCC DAQ device ID, optionally suffixed with ":<board_number>" (default 0)
-DATASET_RID = None  # Replace with your dataset RID.
+DEVICE_ID = "MCC DAQ SERIAL NUMBER:0"  # MCC device unique ID, optionally suffixed with ":<board_number>" (default 0)
+NAME = "mccdaq_test"  # InstroDAQ instance name; prefixes every published channel key
+DATASET_RID = None  # Set to a dataset RID to stream test events to Nominal Core; None to skip publishing.
+
+# Set False to run structure-only checks (finite readings, no value asserts) when the
+# AO->AI loopback harness is not wired.
+LOOPBACK_WIRED = True
 
 # Analog channel mapping
 AI_CH0 = "0"
@@ -187,7 +194,7 @@ class TestMCCDAQHardware(unittest.TestCase):
     def _create_daq(self) -> InstroDAQ:
         """Create, optionally attach publisher, and open a fresh DAQ instance."""
         daq = InstroDAQ(
-            name="mccdaq_test",
+            name=NAME,
             driver=MCCDriver(device_id=DEVICE_ID),
         )
         if DATASET_RID:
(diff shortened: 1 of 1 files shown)


tests/daq/mcc/test_mccdaq_hardware.py
```diff
@@ -42,7 +42,7 @@
 
   Before running, set the following environment variables:
```
hoehner: Total nit but do we want to just say "hand edit the constants below" and not call these "environment variables" ?This could be misleading to someone thinking we're reading from a .env file somewhere.

tests/daq/mcc/test_mccdaq_hardware.py
hoehner: We might need to reorder our tests a bit here. The 'safe state' is now followed up by other tests. Maybe put this at the end?

hoehner (dismissed):

nhschwab:
Comments addressed and test cases pass:

image

hoehner (approved):

PR #192, opened 2026-07-01
fix(visa): quiet pyvisa-py fallback on the healthy path

### LINKED ISSUE #191: VISA: pyvisa-py fallback logs distracting warnings on the healthy path
### Problem

When no IVI VISA implementation is installed, `VisaDriver` falls back to the pyvisa-py (`@py`) backend. This is a perfectly healthy path — the repo ships every `@py` optional dependency — yet it produces noise on **every** `open()`:

1. **Our own warning.** `_open_resource_manager` logs `logger.warning("VISA backend @ivi unavailable ...; falling back to @py")`. Because pyvisa does not cache a *failed* `@ivi` load, this fires on every `open()` / reconnect, for every instrument.
2. **pyvisa-py's chatter.** Constructing `ResourceManager("@py")` emits `warnings.warn("GPIB library not found. ...")`, irrelevant unless the resource actually needs GPIB.

Neither indicates a problem with the connection being made (typically a TCPIP SOCKET, which `@py` serves fine).

### Desired behavior

The fallback should be quiet on the healthy path and only surface a message when a resource genuinely cannot be served by `@py`.

### Proposed fix

- Downgrade the fallback log `warning` -> `debug`.
- Muffle pyvisa-py's construction-time warning by scoping `warnings.catch_warnings()` around the `@py` `ResourceManager(...)` construction.
- Only when a resource fails to open *while on the `@py` fallback*, log one actionable warning (install NI-VISA / set `visa_backend` explicitly), preserving the original exception type.

hoehner:
Closes nominal-io/instro#191

## Problem

When no IVI VISA is installed, `VisaDriver` falls back to the pyvisa-py (`@py`) backend — a healthy path, since the repo ships every `@py` dependency. But it produced noise on **every** `open()`:

1. Our own `logger.warning("... falling back to @py")`. pyvisa doesn't cache a failed `@ivi` load, so this fired on every `open()`, for every instrument.
2. gpib_ctypes' `warnings.warn("GPIB library not found. ...")` at `@py` construction — irrelevant unless the resource needs GPIB.

Neither indicates a problem with the connection actually being made (typically a TCPIP SOCKET, which `@py` serves fine).

## Change

* Downgrade the fallback log `warning` -> `debug`.
* Muffle only the GPIB construction-time warning (`filterwarnings(message=".*GPIB.*")`) around the `@py` `ResourceManager(...)` construction; other warnings still propagate.
* When a resource fails to open *while on the* `@py` *fallback*, log one actionable warning (install NI-VISA / set `visa_backend`) — but only when the failure means `@py` can't serve that resource class, not for ordinary failures (device off, wrong IP). Empirically (pyvisa-py 0.8.1), a capability gap surfaces as `ValueError` ("No class registered for ..." / "gpib_ctypes is installed but could not locate the gpib library"), while connection failures surface as plain `Exception`/`VisaIOError`; we gate on `ValueError` plus `VisaIOError(VI_ERROR_LIBRARY_NFOUND)` for robustness. The exception is re-raised **bare** so the original type is preserved (`discover.py` catches `VisaIOError` specifically).

Net effect: silence on the healthy no-IVI path and on ordinary connection failures; one clear line only when a resource genuinely can't be served by `@py`.

## Tests

`tests/lib/test_visa_driver.py`:

* No warning on a successful fallback.
* Warning logged **and** original exception type propagates when `@py` lacks the library (`VI_ERROR_LIBRARY_NFOUND`) or session class (`ValueError`) for the resource.
* No warning for an ordinary connection failure on the fallback.
* No hint when an explicit backend fails (no fallback occurred).
* GPIB construction warning muffled; unrelated warnings still propagate.

Verified end-to-end against the real pyvisa-py stack (IVI load intercepted to force the fallback): GPIB/VXI opens warn once and raise the original error; a wrong-IP TCPIP open is silent.

`just check` and the full unit suite (1015 passed) are green.

### DIFF
--- a/instro/lib/transports/visa.py
+++ b/instro/lib/transports/visa.py
@@ -9,6 +9,7 @@ import logging
 import socket
 import threading
 import typing
+import warnings
 
 import pyvisa
 from pyvisa.constants import InterfaceType
@@ -151,7 +152,7 @@ class VisaDriver:
             # pyvisa caches one ResourceManager per backend and shares it across every
             # driver in the process; closing it would kill all other drivers' sessions.
             # pyvisa closes it via its own atexit handler.
-            rm = _open_resource_manager(cfg.visa_backend)
+            rm, used_py_fallback = _open_resource_manager(cfg.visa_backend)
             inst: pyvisa.resources.MessageBasedResource | None = None
             try:
                 inst = typing.cast(
@@ -163,6 +164,13 @@ class VisaDriver:
                 if inst is not None:
                     with contextlib.suppress(Exception):
                         inst.close()
+                if used_py_fallback:
+                    logger.warning(
+                        "Failed to open %s on the pyvisa-py (@py) fallback backend (no IVI VISA "
+                        "found). If this instrument needs a vendor VISA library, install NI-VISA "
+                        "or set visa_backend explicitly.",
+                        cfg.visa_resource,
+                    )
                 raise
 
             self._inst = inst
@@ -281,20 +289,24 @@ class VisaDriver:
         return self._inst
 
 
-def _open_resource_manager(backend: str | None) -> pyvisa.ResourceManager:
-    """Open a ResourceManager. An unset backend uses ``@ivi`` and falls back to ``@py``; an explicit backend is used as-is."""
+def _open_resource_manager(backend: str | None) -> tuple[pyvisa.ResourceManager, bool]:
+    """Open a ResourceManager, returning (rm, used_py_fallback). Unset backend uses ``@ivi`` and falls back to ``@py``; an explicit backend is used as-is."""
     if backend is not None:
-        return pyvisa.ResourceManager(backend)
+        return pyvisa.ResourceManager(backend), False
     try:
-        return pyvisa.ResourceManager(DEFAULT_VISA_BACKEND)
+        return pyvisa.ResourceManager(DEFAULT_VISA_BACKEND), False
     except (OSError, pyvisa.errors.Error) as exc:
-        logger.warning(
+        logger.debug(
             "VISA backend %s unavailable (%s); falling back to %s",
             DEFAULT_VISA_BACKEND,
             exc,
             FALLBACK_VISA_BACKEND,
         )
-        return pyvisa.ResourceManager(FALLBACK_VISA_BACKEND)
+        # pyvisa-py warns at construction about optional backends it can't load (e.g. GPIB);
+        # irrelevant unless the resource being opened needs one, so muffle it here.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            return pyvisa.ResourceManager(FALLBACK_VISA_BACKEND), True
 
 
 def _configure_resource(
--- a/instro/lib/transports/visa.py
+++ b/instro/lib/transports/visa.py
@@ -12,7 +12,7 @@ import typing
 import warnings
 
 import pyvisa
-from pyvisa.constants import InterfaceType
+from pyvisa.constants import VI_ERROR_LIBRARY_NFOUND, InterfaceType
 from pyvisa.constants import Parity as VisaParity
 
 logger = logging.getLogger(__name__)
@@ -160,15 +160,14 @@ class VisaDriver:
                     rm.open_resource(cfg.visa_resource),
                 )
                 _configure_resource(inst, cfg)
-            except Exception:
+            except Exception as exc:
                 if inst is not None:
                     with contextlib.suppress(Exception):
                         inst.close()
-                if used_py_fallback:
+                if used_py_fallback and _is_missing_backend_error(exc):
                     logger.warning(
-                        "Failed to open %s on the pyvisa-py (@py) fallback backend (no IVI VISA "
-                        "found). If this instrument needs a vendor VISA library, install NI-VISA "
-                        "or set visa_backend explicitly.",
+                        "The pyvisa-py (@py) fallback backend (no IVI VISA found) cannot serve "
+                        "%s. Install NI-VISA or set visa_backend explicitly.",
                         cfg.visa_resource,
                     )
                 raise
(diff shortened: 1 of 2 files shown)


hoehner:
## Evidence

Reproducible with **no hardware and no NI-VISA** — the script below uses the real `pyvisa-py` backend and only stubs `@ivi` to be unavailable, so the `@ivi -> @py` fallback fires deterministically. It captures exactly what a user would see: stdlib `warnings` **and** WARNING-level logs.

<details><summary>demo script</summary>

```python
import logging, warnings
import pyvisa
from instro.lib.transports.visa import VisaDriver, VisaConfig, _open_resource_manager

# Force @ivi "unavailable" so we always hit the @py fallback (any machine, no hardware).
_real_rm = pyvisa.ResourceManager
def _fake_rm(backend=None, *a, **k):
    if backend == "@ivi":
        raise OSError("simulated: no IVI VISA implementation installed")
    return _real_rm(backend, *a, **k)
pyvisa.ResourceManager = _fake_rm

def _capture(fn):
    logger = logging.getLogger("instro.lib.transports.visa")
    records = []
    h = logging.Handler(); h.setLevel(logging.WARNING)
    h.emit = lambda r: records.append(r.getMessage())
    logger.addHandler(h); logger.setLevel(logging.DEBUG)
    try:
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            try: fn(); raised = None
            except Exception as exc: raised = exc
        return [str(x.message) for x in w], records, raised
    finally:
        logger.removeHandler(h)

# BEFORE (baseline): raw pyvisa-py construction lets the GPIB warning through
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always"); _real_rm("@py")
print("baseline warnings:", [str(x.message) for x in w])

# Scenario 1: healthy fallback
print("healthy:", _capture(lambda: _open_resource_manager(None))[:2])
# Scenario 2: resource fails on @py fallback
w2, l2, e2 = _capture(lambda: VisaDriver("GPIB0::1::INSTR").open())
print("impacted:", w2, l2, type(e2).__name__)
# Scenario 3: explicit backend, no fallback => no hint
print("explicit:", _capture(lambda: VisaDriver(VisaConfig("GPIB0::1::INSTR", visa_backend="@py")).open())[:2])
```
</details>

### Output

```
BEFORE (baseline): raw pyvisa-py construction - what the old path let through
stdlib warnings emitted : ['GPIB library not found. Please manually load it using _load_lib(filename). ...']

AFTER - Scenario 1: HEALTHY fallback (@ivi missing, @py serves the resource)
stdlib warnings emitted : []
WARNING-level logs      : []
=> silent on the healthy path

AFTER - Scenario 2: IMPACTED resource fails on @py fallback (GPIB, no lib)
stdlib warnings emitted : []
WARNING-level logs      : ['Failed to open GPIB0::1::INSTR on the pyvisa-py (@py) fallback backend (no IVI VISA found). If this instrument needs a vendor VISA library, install NI-VISA or set visa_backend explicitly.']
exception type preserved: builtins.ValueError

AFTER - Scenario 3: EXPLICIT backend fails (no fallback => no hint)
stdlib warnings emitted : []
WARNING-level logs      : []
=> no fallback hint (user chose @py explicitly)
```

**Reading:** the old path leaked pyvisa-py's `GPIB library not found` warning on construction; the new path is fully silent on the healthy fallback (S1), emits exactly one actionable line only when a resource actually can't be served by `@py` (S2) while preserving the original exception type, and stays quiet when the user picked `@py` explicitly (S3).

### Targeted tests

```
tests/lib/test_visa_driver.py::test_open_falls_back_to_py_when_ivi_library_not_found PASSED
tests/lib/test_visa_driver.py::test_open_warns_when_resource_fails_on_py_fallback PASSED
tests/lib/test_visa_driver.py::test_open_does_not_warn_when_explicit_backend_fails PASSED
tests/lib/test_visa_driver.py::test_open_does_not_fall_back_for_explicit_backend[@ivi] PASSED
tests/lib/test_visa_driver.py::test_open_does_not_fall_back_for_explicit_backend[@py] PASSED
tests/lib/test_visa_driver.py::test_open_does_not_fall_back_for_explicit_backend[@sim] PASSED
6 passed, 38 deselected
```

instro/lib/transports/visa.py
```diff
@@ -281,20 +289,24 @@ def _require_open_locked(self) -> pyvisa.resources.MessageBasedResource:
...
             "VISA backend %s unavailable (%s); falling back to %s",
             DEFAULT_VISA_BACKEND,
             exc,
             FALLBACK_VISA_BACKEND,
         )
-        return pyvisa.ResourceManager(FALLBACK_VISA_BACKEND)
+        # pyvisa-py warns at construction about optional backends it can't load (e.g. GPIB);
+        # irrelevant unless the resource being opened needs one, so muffle it here.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
nhschwab: This will filter every warning raised during the resource manager construction. We can filter down to just GPIB (or optional dependencies)

`warnings.filterwarnings("ignore", message=".*GPIB.*")`

instro/lib/transports/visa.py
```diff
@@ -163,6 +164,13 @@ def open(self) -> None:
                 if inst is not None:
                     with contextlib.suppress(Exception):
                         inst.close()
+                if used_py_fallback:
+                    logger.warning(
+                        "Failed to open %s on the pyvisa-py (@py) fallback backend (no IVI VISA "
+                        "found). If this instrument needs a vendor VISA library, install NI-VISA "
+                        "or set visa_backend explicitly.",
+                        cfg.visa_resource,
+                    )
```
nhschwab: Every time we fail to open an instrument when the fallback happens, this is the warning that is raised to the user. Even something trivial like the instrument is turned off, a loose cable, wrong visa resource string, will give this warning. 

We should gate on `used_py_fallback` and if the exception is for a missing backend (`VI_ERROR_LIBRARY_NFOUND`)

instro/lib/transports/visa.py
hoehner: Done in d122d0c — narrowed to `filterwarnings("ignore", message=".*GPIB.*")`. Verified against the real stack: the only construction-time warning today is gpib_ctypes' "GPIB library not found" UserWarning, and a test now asserts unrelated warnings still propagate.

instro/lib/transports/visa.py
hoehner: Done in d122d0c — now gated on `used_py_fallback and _is_missing_backend_error(exc)`. One wrinkle worth noting: pyvisa-py doesn't actually raise `VI_ERROR_LIBRARY_NFOUND` here. Probed empirically (pyvisa-py 0.8.1): a GPIB open with the C library missing raises `ValueError("gpib_ctypes is installed but could not locate the gpib library...")`, an unsupported class raises `ValueError("No class registered for ...")`, and a wrong-IP/device-off open raises a plain `Exception("could not connect: ...")`. So the gate is `ValueError` (pyvisa-py's actual signal for a capability gap) plus `VisaIOError(VI_ERROR_LIBRARY_NFOUND)` for robustness. Ordinary failures (device off, loose cable, wrong resource string) match neither and stay silent — verified end-to-end against the real stack with the IVI load intercepted.

instro/lib/transports/visa.py
```diff
@@ -281,20 +288,35 @@ def _require_open_locked(self) -> pyvisa.resources.MessageBasedResource:
...
+
+
+def _is_missing_backend_error(exc: BaseException) -> bool:
+    """True when an open failure means the backend can't serve the resource class, not that the device is unreachable."""
+    if isinstance(exc, pyvisa.errors.VisaIOError):
+        return exc.error_code == VI_ERROR_LIBRARY_NFOUND
+    # pyvisa-py signals "no session class for this resource" (unsupported class, or its
+    # optional library is missing, e.g. GPIB) as ValueError; ordinary connection
+    # failures surface as VisaIOError/OSError/plain Exception.
+    return isinstance(exc, ValueError)
```
nhschwab:
```suggestion
    return isinstance(exc, ValueError) and "No class registered" in str(exc)
```

This avoids false positives for bad visa resource strings and bad serial configs

nhschwab (approved):

PR #224, opened 2026-07-06
fix: close #218 by simplifying scope.mdx language

### LINKED ISSUE #218: docs(guides): cut scope page down to what the library does
User feedback called `scope.mdx` patronizing, and the text bears it out: a "Your Responsibilities" section, "does not absolve you," "You remain responsible," "cannot compensate for misuse." The fix is deletion, not rewording -- most of the page enumerates reader duties true of every software library.

**Delete:** the "What It Doesn't Do" litany, "Your Responsibilities" and subsections, "Validation and Testing," "Implementation Approach" (also oversells compiled binaries -- the default VISA path is pure-Python pyvisa-py), and the summary lecture.

**Keep (about 1/3 of current length):** what instro is for (unified vendor-agnostic interfaces + built-in Nominal publishers), the real-time Warning (scripting/automation, not control loops -- use hardware timing for that), and the native-interface escape hatch (instro is optional; go direct when it serves better; file issues for missing features).

Retitle the frontmatter away from "what responsibilities remain with you as the engineer."

smithednominal:
## Summary

Close #218 - improve scope.mdx (scope of instro, not scopes)

## Type of change

- [ ] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [x] Documentation (`docs`)

## Verification

## Tests

- [ ] Unit tests added or updated
- [x] Existing tests cover this change
- [ ] No tests — explain why:

## Checklist

### DIFF
--- a/docs/guides/instrumentation/scope.mdx
+++ b/docs/guides/instrumentation/scope.mdx
@@ -7,11 +7,12 @@ description: What instro is for, and when to use native interfaces instead
 
 ## What instro provides
 
-- **Unified interfaces and code reusability** across different vendors and models of the same equipment type. For example, if you have a data acquisition system from National Instruments or LabJack, the library provides a consistent API so your code can work with either vendor by changing a configuration parameter.
-- **Common tools and extensions** to the core instrument model
-  - **Background threads** which automatically poll data from instruments
-  - **Built-in publishers** that stream measurements to Nominal Core and Nominal Connect
-  - **Measurement cacheing** which stores the latest measurements for easy access
+- **Unified interfaces and code reusability** across different vendors and models of the same equipment type. For example, switching out an NI DAQmx device for a LabJack device entails switching out the underlying `driver` object while the rest of the code is unchanged.
+- **Background threads** which automatically poll data from instruments on a user-defined cycle or can be overridden to provide application-specific behavior
+- **Publishing classes** to forward `Measurement` data read from instruments or `Command` data sent to instruments on to any number of downstream services, including:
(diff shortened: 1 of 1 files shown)


hoehner (dismissed):

docs/guides/instrumentation/scope.mdx
```diff
@@ -1,104 +1,29 @@
...
-`instro` is primarily used for **scripting and automating test equipment**. It provides:
-
-- **Unified interfaces and code reusability** across different vendors and models of the same equipment type
-- **Built-in publishers** that stream measurements to Nominal Core and Nominal Connect
-
-The library consolidates many software features into one package, making it easier to work with test equipment while maintaining flexibility.
+- **Unified interfaces and code reusability** across different vendors and models of the same equipment type. For example, if you have a data acquisition system from National Instruments or LabJack, the library provides a consistent API so your code can work with either vendor by changing a configuration parameter.
+- **Common tools and extensions** to the core instrument model
+  - **Background threads** which automatically poll data from instruments
+  - **Built-in publishers** that stream measurements to Nominal Core and Nominal Connect
```
hoehner: Can we expand this? Don't want it to look like these are the only two things we can publish to - we should mention the custom publishers as well as the file publishers

smithednominal: ah yeah I was trying to do that with the "common tools and extensions" bit but I can def make more explicit.

hoehner (approved):

PR #226, opened 2026-07-06
feat(cli): report backend and interface coverage in discover

### LINKED ISSUE #210: feat(cli): report backend and interface coverage in discover
`instro discover` prints a raw gpib_ctypes UserWarning mid-banner when no native GPIB library is present. The existing `catch_warnings` at `instro/cli/discover.py:60` only wraps `rm.list_resources()`; the warning fires earlier, at `pyvisa.ResourceManager("@py")` construction (line 57). Meanwhile the `@ivi` -> `@py` fallback is completely silent, so users never learn which backend actually scanned.

Rather than only suppressing, make discover report its own coverage:

1. Widen warning suppression to cover ResourceManager construction on both paths.
2. Print the active backend as a dim line (`@ivi` NI-VISA vs `@py` pyvisa-py, noting the fallback when `@ivi` is absent).
3. On `@py`, read `rm.visalib.get_debug_info()` (verified: returns per-interface status; degraded entries carry the reason) and render each non-Available interface as a dim one-liner with an install hint, e.g. `GPIB: unavailable -- native GPIB library not found (install NI-488.2 or linux-gpib)`.
4. Mention degraded interfaces inside the NO DEVICES FOUND panel, where a missing library most plausibly explains the empty result.

Informative, not an error: the scan continues and the exit code stays 0.

Prior art: nominal-io/instro#191/nominal-io/instro#192 fixed the same warning on the library path.

nhschwab:
Closes #210.

`instro discover` printed a raw gpib_ctypes UserWarning mid-banner when no native GPIB library was present (the existing suppression only wrapped `list_resources()`, but the warning fires at `ResourceManager("@py")` construction), and the `@ivi` → `@py` fallback was completely silent.

## Changes

- **Consolidated ResourceManager construction** on `instro.lib.transports.visa._open_resource_manager` (prior art: #192). The GPIB warning filter now covers all construction paths — including an explicit `visa_backend="@py"` on the library path, which previously still leaked — and the helper returns the active backend. The discover CLI depends on it instead of duplicating it.
- **Backend line**: discover prints the active backend dim under the scan header — `@ivi (system IVI VISA)`, `@py (pyvisa-py)`, or `@py (pyvisa-py — no IVI VISA found)` when the fallback fired.
- **Interface coverage on `@py`**: reads `rm.visalib.get_debug_info()` and renders each non-Available interface family as a dim one-liner with an install hint, e.g. `GPIB: unavailable — gpib_ctypes is installed but could not locate the gpib library (install NI-488.2 or linux-gpib)`.
- **NO DEVICES FOUND panel** repeats the degraded interfaces, where a missing library most plausibly explains the empty result.

Informative, not an error: the scan continues and the exit code stays 0.

## Testing

- 5 new mocked CLI tests (backend line, degraded rendering on fallback and explicit `@py`, panel repetition, construction-time warning suppression); existing visa driver warning-scoping tests pass unchanged.
- Verified end-to-end on real paths: `@ivi` and `@py` natively on macOS against a Tektronix MSO24 over USB (scope found on both backends, GPIB coverage line shown on `@py`, no warning leak), and the no-IVI fallback in a `python:3.12-slim` container using this branch's wheel (fallback label + coverage line + panel mention, no raw warning).
- `just check` green.

### DIFF
--- a/docs/guides/instrumentation/cli.mdx
+++ b/docs/guides/instrumentation/cli.mdx
@@ -24,8 +24,8 @@ instro discover
 To force a specific VISA backend, pass `--backend`:
 
 ```bash
-uv run instro discover --backend @py
-uv run instro discover --backend @ivi
+uv run instro discover --backend=@py
+uv run instro discover --backend "@ivi"
 ```
 
 <Note>
--- a/docs/guides/instrumentation/cli.mdx
+++ b/docs/guides/instrumentation/cli.mdx
@@ -34,6 +34,8 @@ The `@ivi` backend finds an available IVI-compliant backend on your system e.g. 
 
 ### Output
 
+Before scanning, the command prints which VISA backend is active (`@ivi`, or `@py` when no IVI VISA is installed). On the `@py` backend it also lists any interfaces the scan cannot cover — for example `GPIB: unavailable — gpib_ctypes is installed but could not locate the gpib library (install NI-488.2 or linux-gpib)` — so you know when an empty result reflects a missing library rather than an empty bench. These notes are informational; the scan still runs and the command exits 0.
+
 The command groups results into up to three tables.
 
 **Recognized devices** — instruments with a known driver. The table shows the VISA resource address, instrument category, and the `instro` driver class to use.
@@ -42,7 +44,7 @@ The command groups results into up to three tables.
 
 **Unrecognized devices** — instruments that responded to `*IDN?` but did not match any known driver. The raw IDN response is shown so you can identify the device.
 
-If no devices are found at all, the command prints a single error panel.
+If no devices are found at all, the command prints a single error panel; on the `@py` backend the panel repeats any unavailable interfaces.
 
 ### Using a recognized device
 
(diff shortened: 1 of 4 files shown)


hoehner:
One minor nit, can we populate the "Product" column with something like "unknown" instead of just leaving it blank?
image

hoehner:
We should also update the docs to specify the "=" requirement when specifying a backend. 
This didn't work on Windows "uv run instro discover --backend @py"
This DID work on Windows "uv run instro discover --backend=@py"

This is the error I get when I don't have an = sign
image

nhschwab:
> We should also update the docs to specify the "=" requirement when specifying a backend. This didn't work on Windows "uv run instro discover --backend @py" This DID work on Windows "uv run instro discover --backend=@py"
> 
> This is the error I get when I don't have an = sign image

This is specific to powershell treating @ as the splatting operator. Wrapping the argument in double quotes would also work. Going to add a line in docs to make it clear to users

hoehner (approved):

PR #227, opened 2026-07-06
feat(publishers): add jsonl format and deprecate the array-JSON writer

### LINKED ISSUE #212: feat(publishers): add jsonl format and deprecate the array-JSON writer
`JsonFileWriter.write` (`instro/lib/publishers/files.py:95`) re-reads the whole file, appends one record, and rewrites the entire array with `indent=2` -- per measurement. At daemon polling rates that's O(n^2) I/O. User feedback: "this feels near unusable. How about switching to newline-delimited JSON?"

We're switching, without breaking:

- Add `format="jsonl"` to the `FilePublisher` Literal, backed by a `JsonlFileWriter` that opens its handle once in `__init__` (append text mode, like `AvroFileWriter`) and writes `json.dumps(data.__dict__)` + newline + flush per measurement. O(1), crash-safe (every flushed line is a complete record), `.jsonl` extension.
- Same record shape as the existing JSON writer so consumers switching formats see identical fields.
- `format="json"` keeps working exactly as today but emits a `DeprecationWarning` pointing at `jsonl`; removal deferred to a future major release.
- Docs and examples switch to `jsonl` as the JSON-family recommendation; the publishers guide notes why (`json` rewrites the whole file per measurement).
- Tests alongside existing FilePublisher tests, including `close()` and the deprecation warning.

hoehner:
Closes nominal-io/instro#212

## What

* Adds `format="jsonl"` to `FilePublisher`, backed by a new `JsonlFileWriter` that opens its file handle once in `__init__` (append text mode, like `AvroFileWriter`) and writes one `json.dumps(data.__dict__)` line + flush per publish — O(1) per measurement, and every flushed line is a complete record, so a crash mid-run leaves a readable `.jsonl` file.
* Records carry the exact same fields as the existing array-JSON writer, so consumers switching formats see identical shapes (covered by a test comparing the two writers' output).
* `format="json"` keeps working exactly as before but now emits a `DeprecationWarning` pointing at `jsonl`; removal is deferred to a future major release.
* Docs: the publishers guide now recommends `jsonl` for human-readable captures and marks `json` as deprecated with the O(n^2) rationale; the overview page mentions JSON Lines.

## Testing

* New `tests/lib/test_file_publisher.py`: line-per-record output, record-shape parity with the JSON writer, per-publish flush (file complete before `close()`), idempotent `close()`, `DeprecationWarning` on `format="json"`, no warning on `jsonl`, and the legacy array format still working.
* `just check` and the full unit suite (1022 passed) are green.

### DIFF
--- a/instro/lib/publishers/files.py
+++ b/instro/lib/publishers/files.py
@@ -2,6 +2,7 @@
 
 import csv
 import json
+import math
 import warnings
 from datetime import datetime
 from pathlib import Path
@@ -133,17 +134,31 @@ class JsonFileWriter:
         pass
 
 
+def _json_safe(value: Any) -> Any:
+    if isinstance(value, float) and not math.isfinite(value):
+        return None
+    if isinstance(value, dict):
+        return {k: _json_safe(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [_json_safe(v) for v in value]
+    return value
+
+
 class JsonlFileWriter:
     """Handles newline-delimited JSON writing with a persistent file handle."""
 
     def __init__(self, file_path: Path):
         self.file_path = file_path
-        self.file_path.parent.mkdir(parents=True, exist_ok=True)
+        self._ensure_file_exists()
         self._file = open(self.file_path, "a")
 
+    def _ensure_file_exists(self):
+        """Create directory if it doesn't exist."""
+        self.file_path.parent.mkdir(parents=True, exist_ok=True)
+
     def write(self, data: Measurement | Command):
-        """Append data as one JSON line."""
-        self._file.write(json.dumps(data.__dict__) + "\n")
+        """Append data as one JSON line, mapping non-finite floats to null."""
+        self._file.write(json.dumps(_json_safe(data.__dict__), allow_nan=False) + "\n")
         self._file.flush()
 
     def open(self):
@@ -151,8 +166,7 @@ class JsonlFileWriter:
 
     def close(self):
         """Close the file writer."""
-        if self._file and not self._file.closed:
-            self._file.close()
+        self._file.close()
 
 
 class CsvFileWriter:
--- a/instro/lib/publishers/files.py
+++ b/instro/lib/publishers/files.py
@@ -1,7 +1,8 @@
-"""File-based publishers (JSON, CSV, Avro)."""
+"""File-based publishers (JSON, JSONL, CSV, Avro)."""
 
 import csv
 import json
+import warnings
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Literal, Protocol
@@ -25,14 +26,14 @@ class FilePublisher:
     def __init__(
         self,
         directory: str | Path,
-        format: Literal["json", "csv", "avro"] = "avro",
+        format: Literal["json", "jsonl", "csv", "avro"] = "avro",
         custom_file_name: str | None = None,
     ):
         """Write Measurement/Command data to a file.
 
         Args:
             directory: Output directory.
-            format: ``"json"``, ``"csv"``, or ``"avro"`` (default).
+            format: ``"jsonl"``, ``"csv"``, ``"avro"`` (default), or ``"json"`` (deprecated).
             custom_file_name: Filename without extension; defaults to ``measurements-<UTC-timestamp>``.
         """
         self.directory = Path(directory)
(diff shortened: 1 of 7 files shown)


instro/lib/publishers/files.py
```diff
@@ -124,6 +133,28 @@ def close(self):
         pass
 
 
+class JsonlFileWriter:
+    """Handles newline-delimited JSON writing with a persistent file handle."""
+
+    def __init__(self, file_path: Path):
+        self.file_path = file_path
+        self.file_path.parent.mkdir(parents=True, exist_ok=True)
+        self._file = open(self.file_path, "a")
```
nhschwab, line 153: Can we split this out into another method `def _ensure_file_exists(self)` to match precedent set by other file publishers?

instro/lib/publishers/files.py
```diff
@@ -124,6 +133,28 @@ def close(self):
...
+        self._file.write(json.dumps(data.__dict__) + "\n")
+        self._file.flush()
+
+    def open(self):
+        pass
+
+    def close(self):
+        """Close the file writer."""
+        if self._file and not self._file.closed:
+            self._file.close()
```
nhschwab: If a publisher is shared by multiple instruments, when the first instrument is closed, it also closes the publisher, which means the jsonl file is closed while the second instrument may still be trying to write data to the jsonl file. This will result an IO error. The avro publisher is also prone to this error (CSV and deprecated JSON are tolerant since they do not have handlers). Up to you if we need to fix that in this PR, but we should definitely open a new issue to fix that since it affects the avro publisher and this new jsonl publisher

instro/lib/publishers/files.py
```diff
@@ -124,6 +133,28 @@ def close(self):
...
+    """Handles newline-delimited JSON writing with a persistent file handle."""
+
+    def __init__(self, file_path: Path):
+        self.file_path = file_path
+        self.file_path.parent.mkdir(parents=True, exist_ok=True)
+        self._file = open(self.file_path, "a")
+
+    def write(self, data: Measurement | Command):
+        """Append data as one JSON line."""
+        self._file.write(json.dumps(data.__dict__) + "\n")
```
nhschwab: `json.dumps()` has `allow_nan=True` by default, and `InstroScope` has drivers that explicitly produce NaN measurements. This won't error, but it will mean that most parsers that try to ingest the jsonl file output will fail. We should include a mapping to null instead

```import math

  def _json_safe(value: Any) -> Any:
      if isinstance(value, float) and not math.isfinite(value):
          return None
      if isinstance(value, dict):
          return {k: _json_safe(v) for k, v in value.items()}
      if isinstance(value, list):
          return [_json_safe(v) for v in value]
      return value
```


```suggestion
        self._file.write(json.dumps(_json_safe(data.__dict__), allow_nan=False) + "\n")
```

instro/lib/publishers/files.py
nhschwab:
```suggestion
        self._file.close()
```

`file.close()` is idempotent

nhschwab:
Out of scope for this PR but we should cut a new issue if not fixed here:

In `instro/lib/publishers/publisher.py`, `BufferedPublisher` maintains stale measurements in the buffer after it is closed, so a secondary close call rewrites stale measurements to the endpoints. This also triggers the already mentioned bug where calling close on the already closed jsonl and avro handlers raises an error.

instro/lib/publishers/files.py
hoehner: Good catch, kept this PR scoped and opened #231 to track the shared-publisher lifecycle for both the avro and jsonl writers.

instro/lib/publishers/files.py
hoehner: Done in c3c5459, kept the `open()` call in `__init__` to match the `AvroFileWriter` shape, with `_ensure_file_exists` doing the mkdir.

instro/lib/publishers/files.py
hoehner: Done in c3c5459, with a test covering NaN/±inf mapping to null. Left the deprecated array-JSON writer as-is since it's frozen for removal.

instro/lib/publishers/files.py
hoehner: Done in c3c5459. (Left the equivalent guard in `AvroFileWriter` alone. There it protects the `flush()` call, which raises on a closed file.)

nhschwab (dismissed):

nhschwab (approved):

hoehner:
> Out of scope for this PR but we should cut a new issue if not fixed here:
> 
> In `instro/lib/publishers/publisher.py`, `BufferedPublisher` maintains stale measurements in the buffer after it is closed, so a secondary close call rewrites stale measurements to the endpoints. This also triggers the already mentioned bug where calling close on the already closed jsonl and avro handlers raises an error.

@nhschwab made this issue here: https://github.com/nominal-io/instro/issues/232

PR #229, opened 2026-07-06
docs: batch clarity fixes from user feedback

### LINKED ISSUE #217: docs: batch clarity fixes from user feedback
Five small fixes from external user feedback, one PR:

**(a) Connect cross-links.** `installation.mdx` "Using Nominal Connect" and the publishers page's `NominalConnectPublisher` section don't reference each other. One link each direction.

**(b) Custom-publisher guide.** Link `Measurement`/`Command` to their reference entries at first mention. Add a paragraph answering "why isn't publishing async by default?": the channel buffer is itself a publisher, and `get_channel`'s blocking semantics require samples in-buffer when the daemon iteration completes; `QueuedPublisher` is the opt-in wrapper for slow sinks.

**(c) Quickstart teardown rationale.** One sentence at the teardown step: output-off before close is bench safety; `close()` tears down the daemon and flushes publishers, not just the socket.

**(d) Extension points.** State in `custom-instruments.mdx` that `_package_command`/`_package_measurement` are stable, supported extension points for subclass authors (their docstrings already address that audience; no rename -- protected-method usage is standard Python). The reference site filters `!^_`, so add a per-page mkdocstrings `filters` override on `reference/instrument.md` whitelisting these two so the reference matches the guide.

**(e) Quickstart: one or two more calls.** Show `set_ovp_level` + `ovp_enable` next to `set_current_limit` -- protection fits the existing step and the simulator models trip behavior. The quickstart otherwise stays as-is.

hoehner:
Closes nominal-io/instro#217. Five small fixes from external user feedback:

**(a) Connect cross-links.** `installation.mdx` "Using Nominal Connect" now links to the publishers page's `NominalConnectPublisher` section, and that section links back to the install instructions.

**(b) Custom-publisher guide.** `Measurement`/`Command` link to their reference entries at first mention. Added a paragraph after the Publisher protocol explaining why `publish` is synchronous by default (the channel buffer behind `get_channel` is itself a publisher, and `get_channel`'s blocking semantics need samples in-buffer when the daemon iteration completes) and pointing at `QueuedPublisher` as the opt-in wrapper for slow sinks.

**(c) Quickstart teardown rationale.** One sentence at the teardown step: output-off before disconnect is bench safety; `close()` stops the daemon and flushes publishers, not just the socket.

**(d) Extension points.** `custom-instruments.mdx` now states `_package_command`/`_package_measurement` are stable, supported extension points for subclass authors. Added a per-page mkdocstrings `filters` override on `reference/instrument.md` whitelisting the two methods so the reference page matches. Verified in a local `mkdocs build`: exactly those two underscored members render, nothing else leaks.

**(e) Quickstart protection calls.** `set_overvoltage_protection_level` + `set_overvoltage_protection_enabled` next to `set_current_limit` (the issue's `set_ovp_level`/`ovp_enable` shorthand mapped to the real HAL names). The simulator models trip behavior, noted in the step.

`mint broken-links` flags none of the new links (the 25 pre-existing hits are Windows path-separator false positives on links to pages that exist).

### DIFF
--- a/docs/guides/instrumentation/publishers.mdx
+++ b/docs/guides/instrumentation/publishers.mdx
@@ -68,8 +68,8 @@ class Publisher(Protocol):
 
 ### Data Types
 
-- **`Measurement`**: Contains data read from the instrument (e.g., analog input readings, voltage measurements). Includes channel data, timestamps, and optional tags.
-- **`Command`**: Represents a command sent to the instrument (e.g., setting voltage, enabling output). Includes channel data, a single timestamp, and optional tags.
+- [**`Measurement`**](https://nominal-io.github.io/instro/reference/types/#instro.lib.types.Measurement): Contains data read from the instrument (e.g., analog input readings, voltage measurements). Includes channel data, timestamps, and optional tags.
+- [**`Command`**](https://nominal-io.github.io/instro/reference/types/#instro.lib.types.Command): Represents a command sent to the instrument (e.g., setting voltage, enabling output). Includes channel data, a single timestamp, and optional tags.
 
 ## Built-in Publishers
 
@@ -114,7 +114,7 @@ The `NominalCorePublisher` includes built-in batching via the Nominal Core Pytho
 
 ### NominalConnectPublisher
 
-Streams real-time data to Nominal Connect for live visualization and monitoring during tests that use the Nominal Connect Desktop Application.
+Streams real-time data to Nominal Connect for live visualization and monitoring during tests that use the Nominal Connect Desktop Application. Nominal Connect apps can also install `instro` for you; see [Using Nominal Connect](/instrumentation/installation#using-nominal-connect).
 
 **Constructor Parameters:**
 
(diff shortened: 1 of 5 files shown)


docs/guides/instrumentation/publishers.mdx
```diff
@@ -263,6 +263,8 @@ class Publisher(Protocol):
     def close(self): ...
 ```
 
+`publish` runs synchronously, on the same thread as the instrument method that produced the data. That is deliberate: the in-memory channel buffer behind `get_channel` is itself a publisher, and `get_channel`'s blocking semantics rely on samples being in the buffer by the time each background-daemon iteration completes. If publishing were deferred to a background thread by default, a `get_channel` call could miss a sample the daemon just collected. Keep custom `publish` implementations fast; when the sink is slow (network, disk), wrap the publisher in a [`QueuedPublisher`](#queuedpublisher) to opt into background publishing.
+
```
cartercanedy: This is a pretty dense bit of documentation. It makes sense when framed as the classic buffered vs unbuffered io debate. I'd make that framing clearer, and reorganize this underneath the subclassing example and coalesce the parent class declaration example code block with the example derivation code block, and put this paragraph under a new subheading like "Publisher execution model" or "Publisher data handling model" (or something like that).

cartercanedy (approved):

Issue #232, opened 2026-07-06
fix(publishers): BufferedPublisher retains its buffer after close, replaying stale records on double close


nhschwab:
## Problem

`BufferedPublisher.close()` calls `publish_batch()` and then closes the wrapped publisher, but never clears `self.buffer` (`instro/lib/publishers/publisher.py:36-38`) — `publish_batch()` iterates the buffer without clearing it (`:42-44`), and `publish()` only clears on the size threshold (`:28-30`). `Instrument.close()` has no closed flag and `__exit__` calls `close()` unconditionally, so an explicit `close()` inside a `with` block produces a double close.

On the second close the stale buffer is replayed into the wrapped publisher:

- with handle-based writers (`avro`, `jsonl`): `ValueError: I/O operation on closed file`, raised out of `__exit__`
- with `csv` / deprecated `json`: silent duplicate records

## Repro

```python
pub = BasicBufferedPublisher(FilePublisher(tmp_dir, format="jsonl"), buffer_size=1000)
pub.publish(measurement)   # buffered, below threshold
pub.close()                # drains batch, closes file; buffer still holds the record
pub.close()                # replays buffer -> ValueError: I/O operation on closed file
```

## Possible fixes

- `publish_batch()` clears the buffer after draining (flush semantics)
- `Instrument.close()` made idempotent

Distinct from #231 (shared-publisher handle lifecycle): this one is reachable with a single instrument. Raised in review of #227.

Issue #231, opened 2026-07-06
FilePublisher shared across instruments breaks handle-based writers (avro, jsonl, csv)


hoehner:
## Problem

When one `FilePublisher` is shared by multiple instruments, the first instrument to close also closes the publisher, which closes the underlying file handle. Any subsequent `publish()` from another instrument still using that publisher raises `ValueError: I/O operation on closed file`.

This affects the writers that hold a persistent file handle:

- `AvroFileWriter` (pre-existing)
- `JsonlFileWriter` (added in #227)
- `CsvFileWriter` (holds a persistent handle since #233)

Only the deprecated `JsonFileWriter` is tolerant, because it reopens the file on every write. All non-deprecated formats are now affected.

## Plan: lazy reopen on publish

Each handle-based writer in `instro/lib/publishers/files.py` will reopen its file in append mode when a `publish()` arrives after `close()`:

- `close()` keeps its current behavior: flush and close the handle. Instruments keep closing their publishers for free.
- `write()` on a closed writer reopens the file in append mode and continues. Shared-publisher sequence: instrument A closes → file closes; instrument B publishes → reopens in append; instrument B closes → file closes for good.
- `AvroFileWriter` reopens with `"a+b"` so fastavro appends to the existing container (its `__init__` uses `"wb"`, which truncates — reopen must not). Verify with a close→publish→close round-trip test that the file stays a valid Avro container.
- `CsvFileWriter` must not rewrite the header on reopen; `JsonlFileWriter` reopens with plain `"a"`.
- Log on reopen so a publish after an intentional shutdown stays observable instead of silently resurrecting the file.

Everything stays inside `files.py` — no `Publisher` protocol change, no instrument changes, and wrapped shared publishers (`QueuedPublisher`/`BufferedPublisher` around a `FilePublisher`) are fixed with no wrapper changes because the fix lives on the publish path.

## Directions considered and rejected

- **Caller-owned lifecycle** (HALs don't close publishers they didn't create): breaks the close-for-free UX for the common single-instrument case.
- **Reference-counting via `open()` on attach**: the `Publisher` protocol has no `open()` — adding one is a breaking change to a structural protocol, and `BufferedPublisher`/`QueuedPublisher` don't forward it, so counts go unbalanced and the bug returns.
[...]

PR #233, opened 2026-07-06
refactor(publishers): hold a persistent file handle in CsvFileWriter

### LINKED ISSUE #213: refactor(publishers): hold a persistent file handle in CsvFileWriter
`CsvFileWriter.write` (`instro/lib/publishers/files.py:140`) reopens the file on every write. Not O(n^2) like the JSON writer, but the same missing-held-handle pattern the stubbed `open()` methods were always meant to fix. Open the handle once (as `AvroFileWriter` does and the new `JsonlFileWriter` will), write/flush per measurement, close in `close()`.

Companion to nominal-io/instro#212; either PR order works.

nhschwab:
Closes #213.

`CsvFileWriter.write` reopened the file on every publish (`exists()` check + `open` per call). This moves it to the `AvroFileWriter` shape the stubbed `open()` was always anticipating:

- Handle opened once in `__init__` (`"a"` mode, `newline=""`), one `csv.writer` bound to it
- `write()` emits rows and flushes per publish; `close()` closes the handle (bare `close()`; file close is idempotent, per the convention settled in #227 review)
- Behavior otherwise preserved: `"a"` creates-or-appends exactly like the old `"a" if exists else "w"`, and the per-instance header flag is untouched (append-across-instances still re-emits the header, asserted in a test so any future change to that is deliberate)

## Known interactions (deferred by design)

- **#231** (shared-publisher close lifecycle): CSV now joins avro/jsonl in the persistent-handle column — publish-after-close on a shared publisher raises instead of silently reopening. #231's note that "CsvFileWriter is tolerant because it reopens per write" stops being true once this merges.
- **#232** (BufferedPublisher stale-buffer replay on double close): the replay now hits a closed CSV handle too. Not addressed here.
- **#227** (jsonl): companion PR; either merge order works. Tests live in a new `tests/lib/test_csv_file_writer.py` specifically to avoid an add/add conflict with #227's `test_file_publisher.py`.

## Testing

- 4 new tests: header-once + one row per value (Measurement and Command), rows readable on disk before `close()` (per-publish flush), single handle across writes with idempotent close, and append-across-instances semantics preserved.
- Full `tests/lib` suite (62) and `just check` green.

### DIFF
--- a/instro/lib/publishers/files.py
+++ b/instro/lib/publishers/files.py
@@ -125,29 +125,26 @@ class JsonFileWriter:
 
 
 class CsvFileWriter:
-    """Handles CSV format writing with proper file management."""
+    """Handles CSV format writing with a persistent file handle."""
 
     def __init__(self, file_path: Path):
         self.file_path = file_path
         self._headers_written = False
         self._ensure_file_exists()
+        self._file = open(self.file_path, "a", newline="")
+        self._writer = csv.writer(self._file)
 
     def _ensure_file_exists(self):
-        """Create directory and file if they don't exist."""
+        """Create directory if it doesn't exist."""
         self.file_path.parent.mkdir(parents=True, exist_ok=True)
-        # CSV file will be created when first write occurs
 
     def write(self, data: Measurement | Command):
         """Append data to CSV file."""
-        mode = "a" if self.file_path.exists() else "w"
-
-        with open(self.file_path, mode, newline="") as f:
-            writer = csv.writer(f)
-
-            if isinstance(data, Measurement):
-                self._write_measurement(writer, data)
-            elif isinstance(data, Command):
-                self._write_command(writer, data)
+        if isinstance(data, Measurement):
+            self._write_measurement(self._writer, data)
+        elif isinstance(data, Command):
+            self._write_command(self._writer, data)
+        self._file.flush()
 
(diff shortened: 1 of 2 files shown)


instro/lib/publishers/files.py
```diff
@@ -134,65 +124,27 @@ def close(self):
...
     def __init__(self, file_path: Path):
         self.file_path = file_path
         self._headers_written = False
         self._ensure_file_exists()
+        self._file = open(self.file_path, "a", newline="")
+        self._writer = csv.writer(self._file)
 
     def _ensure_file_exists(self):
-        """Create directory and file if they don't exist."""
+        """Create directory if it doesn't exist."""
```
hoehner, line 185: Should we migrate the writing of the header rows to this point immediately after the file is created? Seems silly to defer that to the other methods since row writing should only ever happen one time.

hoehner (changes requested):

stefanv:
I don't have push permission to this branch, but made the requested changes in https://github.com/nominal-io/instro/pull/236 (PR onto this PR).

nhschwab:
> I don't have push permission to this branch, but made the requested changes in #236 (PR onto this PR).

Merged your PR and added a one line check for empty file before writing header to avoid header duplication across multiple publishers sharing the same output file. Thanks for the help!

hoehner (approved):

PR #240, opened 2026-07-07
fix(publishers): reopen file writers on publish after close

### LINKED ISSUE #231: FilePublisher shared across instruments breaks handle-based writers (avro, jsonl, csv)

hoehner:
Fixes nominal-io/instro#231.

When one `FilePublisher` is shared by multiple instruments, the first instrument to close also closed the underlying file handle, so any later `publish()` from another instrument raised `ValueError: I/O operation on closed file`. This affected all handle-based writers: `AvroFileWriter`, `JsonlFileWriter`, and `CsvFileWriter` (since nominal-io/instro#233).

## Approach: lazy reopen on publish

Per the design discussion on the issue, each handle-based writer now reopens its file in append mode when a `write()` arrives after `close()`:

* `close()` is unchanged: flush and close. Instruments keep closing their publishers for free.
* `write()` on a closed writer logs the reopen (`logger.info`) and reopens in append mode, then continues.
* `AvroFileWriter` reopens with `"a+b"` so fastavro appends to the existing container (`__init__` keeps `"wb"`). Verified by a close→publish→close round-trip test that reads the file back with `fastavro.reader`.
* `CsvFileWriter` reuses its existing empty-file check on reopen, so the header is never written twice.
* `JsonlFileWriter` reopens with plain `"a"`.

Everything stays inside `files.py`: no `Publisher` protocol change, no instrument changes, and shared publishers wrapped in `QueuedPublisher`/`BufferedPublisher` are fixed with no wrapper changes because the fix lives on the publish path.

## Testing

* New close→publish→close round-trip tests for all three handle-based formats (JSONL, CSV header-once, Avro container validity via `fastavro.reader`).
* `just check` clean; full unit suite passes (1066 passed).

## Docs

Added a "Sharing one file across instruments" note to the `FilePublisher` section of `docs/guides/instrumentation/publishers.mdx`.


instro/lib/publishers/files.py
```diff
@@ -239,18 +256,23 @@ def __init__(self, file_path: Path):
...
 
     def _ensure_file_exists(self):
         """Create directory if it doesn't exist."""
         self.file_path.parent.mkdir(parents=True, exist_ok=True)
 
+    def _open_file(self, mode: Literal["wb", "a+b"]):
+        """Open the file and initialize the Avro writer."""
+        self._file = open(self.file_path, mode)
+        # Using snappy compression requires cramjam package
+        self._writer = fastavro.write.Writer(self._file, self._parsed_schema, codec="snappy")
```
cartercanedy, line 269: From a Devin review:

When `AvroFileWriter._open_file("a+b")` is called after close (`instro/lib/publishers/files.py:275`), a new `fastavro.write.Writer` is constructed on the append-mode file handle. This writes a fresh Avro container header (magic bytes, schema metadata, and a new random sync marker) at the current position, producing a concatenated multi-container file rather than extending the original container's data blocks. The Avro specification permits concatenated containers, and `fastavro.reader` handles them (the test at `tests/lib/test_file_publisher.py:155-156` confirms this). However, not all Avro consumers handle concatenated containers equally well. If downstream tooling (e.g., Apache Spark, certain Java Avro SDK versions, or Nominal Core's ingest pipeline) expects a single-container file, the reopened file could fail to parse. Worth confirming that Nominal Core's Avro ingest path accepts concatenated containers.

instro/lib/publishers/files.py
```diff
@@ -175,18 +185,25 @@ class CsvFileWriter:
...
 
-    def _ensure_file_exists(self):
-        """Create directory if it doesn't exist."""
-        self.file_path.parent.mkdir(parents=True, exist_ok=True)
-
     def write(self, data: Measurement | Command):
         """Append data to CSV file."""
+        if self._file.closed:
+            logger.info("Reopening closed CSV writer for append: %s", self.file_path)
+            self._open_file()
```
cartercanedy, line 206: This is racy: two threads call `publisher.write(...)` at the same time, both are able to observe that `self._file.closed` is true before `self._file` is replaced.

Could be solved easily with a mutex

instro/lib/publishers/files.py
```diff
@@ -239,18 +256,23 @@ def __init__(self, file_path: Path):
...
+        """Open the file and initialize the Avro writer."""
+        self._file = open(self.file_path, mode)
+        # Using snappy compression requires cramjam package
+        self._writer = fastavro.write.Writer(self._file, self._parsed_schema, codec="snappy")
+
     def write(self, data: Measurement | Command):
         """Append data to Avro file."""
+        if self._file.closed:
+            logger.info("Reopening closed Avro writer for append: %s", self.file_path)
+            self._open_file("a+b")
```
cartercanedy, line 275: same race condition is possible here

cartercanedy (changes requested):

hoehner:
Preferring the SharedPublisher implementation documented here instead of this approach.
https://github.com/nominal-io/instro/pull/250

PR #250, opened 2026-07-08
feat(publishers): shared publisher

### LINKED ISSUE #231: FilePublisher shared across instruments breaks handle-based writers (avro, jsonl, csv)

cartercanedy:
## Summary

- Updates docs to clarify that publishers are exclusive resources owned by instruments, with lifetimes coupled to the owning instrument's lifetime.
- Adds `SharedPublisher` as an explicit escape hatch for sharing one underlying publisher across multiple instruments via reference counting.

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

- Added tests to verify the mechanics of the shared publisher class

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

- New `SharedPublisher` internals in `publisher.py` should be scrutinized most


docs/guides/instrumentation/publishers.mdx
```diff
@@ -294,7 +303,44 @@ Keep custom `publish` implementations fast. When the sink is slow (network, disk
 
 ## Publisher Wrappers
 
-Publisher wrappers modify the behavior of other publishers by adding buffering or asynchronous processing. They're composable - wrap any publisher to add new capabilities.
+Publisher wrappers modify the behavior of other publishers by adding buffering, asynchronous processing, or explicit shared ownership. They are composable: wrap any publisher to add new capabilities.
+
+### SharedPublisher
+
+`SharedPublisher` is an escape hatch for sharing one underlying publisher across multiple instruments. The default model is still exclusive ownership: one publisher instance belongs to one instrument. Use `SharedPublisher` only when multiple instruments must publish to the same file, stream, client, or other sink.
```
benliepert: I'd avoid the term escape hatch, it makes it sound hacky when AFAICT this is a proper feature

examples/publishers/publish_shared.py
```diff
@@ -0,0 +1,39 @@
+"""Example: Publishers: publish shared."""
+
+from instro.lib.publishers import FilePublisher
+from instro.lib.publishers.publisher import SharedPublisher
```
benliepert: should be able to import from `instro.lib.publishers`

docs/guides/instrumentation/publishers.mdx
```diff
@@ -294,7 +303,44 @@ Keep custom `publish` implementations fast. When the sink is slow (network, disk
...
+SharedPublisher(
+    publisher: Publisher  # The underlying publisher to share
+)
+```
+
+**Example:**
+
+```python
+from instro.lib.publishers import FilePublisher
+from instro.lib.publishers.publisher import SharedPublisher
```
benliepert: should be able to import from `instro.lib.publishers`

instro/lib/publishers/publisher.py
```diff
@@ -81,3 +80,85 @@ def close(self):
...
+        with self.__state_lock:
+            if (state := self.__state) is None:
+                raise RuntimeError("attempted to publish to a shared publisher handle that was already closed")
+            state.publish(data, **kwargs)
+
+    def close(self) -> None:
+        """Close this instance and possibly release the underlying publisher."""
+        with self.__state_lock:
+            if self.__state is not None:
+                self.__state.decrement()
```
benliepert: if `decrement()` calls `close()` and that throws, `self.__state` won't ever be set to `None`

repro:
```python
from instro.lib import Measurement
from instro.lib.publishers import SharedPublisher


class FailingClosePublisher:
    def publish(self, data, **kwargs):
        print("published")

    def close(self):
        raise OSError("flush failed")


shared = SharedPublisher(FailingClosePublisher())

try:
    shared.close()
except OSError as exc:
    print("first close:", exc)

try:
    shared.close()
except Exception as exc:
    print("retry close:", type(exc).__name__, exc)

shared.publish(Measurement(channel_data={"v": [1.0]}, timestamps=[1]));
```

oneliner:
```
uv run python -c 'from instro.lib import Measurement; from instro.lib.publishers import SharedPublisher; FailingClosePublisher = type("FailingClosePublisher", (), {"publish": lambda self, data, **kwargs: print("published"), "close": lambda self: (_ for _ in ()).throw(OSError("flush failed"))}); shared = SharedPublisher(FailingClosePublisher()); exec("try:\n    shared.close()\nexcept OSError as exc:\n    print(\"first close:\", exc)\ntry:\n    shared.close()\nexcept Exception as exc:\n    print(\"retry close:\", type(exc).__name__, exc)\nshared.publish(Measurement(channel_data={\"v\": [1.0]}, timestamps=[1]))")'
```

benliepert (dismissed):
see comment

examples/publishers/publish_shared.py
```diff
@@ -0,0 +1,38 @@
...
+    name="primaryPSU",
+    driver=SimulatedPSU(PRIMARY_VISA_RESOURCE),
+    num_channels=2,
+    publishers=[shared_publisher],
+)
+secondary_psu = InstroPSU(
+    name="secondaryPSU",
+    driver=SimulatedPSU(SECONDARY_VISA_RESOURCE),
+    num_channels=2,
+    publishers=[shared_publisher.clone()],
```
hoehner, line 23: The dev-ex of having to remember to call .clone() on the shared publisher feels strange to me. It seems like we shouldn't have to do this by nature of the shared publisher being called the "shared publisher". This is a semantics thing and not a functional rebuke of this approach - it just seems like the kind of thing that users will totally forget to do which makes me question if this is really how we should do it (I don't have better ideas for hooks on when to increment the refcount unfortunately).

cartercanedy: Good docs are a good place to start. IMO it's no different than if we didn't couple the lifetime of publishers to the lifetime of instruments and required the user to call close. This way preserves the ability to write a one-liner while providing that escape hatch when you need to get jiggy with it and increase the complexity of the application

instro/lib/publishers/publisher.py
```diff
@@ -81,3 +80,87 @@ def close(self):
...
+        with self.__state_lock:
+            if (state := self.__state) is None:
+                raise RuntimeError("attempted to clone a shared publisher handle that was already closed")
+            return SharedPublisher.__from_state(state.increment())
+
+    def publish(self, data: Measurement | Command, **kwargs) -> None:
+        """Publish data to the underlying publisher being shared."""
+        with self.__state_lock:
+            if (state := self.__state) is None:
+                raise RuntimeError("attempted to publish to a shared publisher handle that was already closed")
```
hoehner, line 152: How about the case where an instrument starts - publishes to file closes and then ANOTHER instrument starts hoping to use the same publisher?

instro/lib/publishers/publisher.py
```diff
@@ -81,3 +80,87 @@ def close(self):
...
         self._thread.join()
         self.publisher.close()
+
+
+class SharedPublisher(Publisher):
+    """A publisher that can be shared between multiple instruments."""
+
+    class __ControlBlock:
+        def __init__(self, publisher: Publisher):
+            self.__lock = threading.RLock()
```
hoehner: Does this need to be RLock still or could it also be Lock

instro/lib/publishers/publisher.py
```diff
@@ -81,3 +80,87 @@ def close(self):
         self._stop_event.set()
         self._thread.join()
         self.publisher.close()
+
+
+class SharedPublisher(Publisher):
```
hoehner, line 85: In general can we try to stick to the single underscore pattern? Just trying to keep that pattern tight before merging.

hoehner, line 85: Can we also consider updating our error message to guide the user to use the .clone() method if they accidentally forget to use it when passing to a new instrument?

hoehner: This might belong in the FilePublisher class, not the SharedPublisher class

instro/lib/publishers/publisher.py
cartercanedy: you'd just need to clone it again and hold on to another instance to prolong the lifetime

hoehner: Cool that works for me

hoehner (approved):

PR #256, opened 2026-07-09
fix(pkg): add instro/__init__.py so tab completion works

stefanv:
instro had no `__init__.py`, making it an implicit namespace package without attributes until each subpackage was explicitly imported. This meant that `import instro; instro.<TAB>` found no completions in IPython.

Add an `__init__.py` that eagerly imports the always-present subpackages (cli, daq, dmm, etc.).
It uses `pkgutil`'s `extend_path` to keep the namespace packages working.

Needed to be quite careful exactly how `extend_path` was called, because griffe recognizes namespace packages via regex (!).

I suspect many engineers will want to explore this package from the Python terminal / IPython, so I feel like this is an important usability improvement.

### DIFF
--- a/instro/__init__.py
+++ b/instro/__init__.py
@@ -1,0 +1,13 @@
+"""Typed Python API for test-and-measurement instruments."""
+
+# Let workspace packages (instro-contrib, instro-unstable, instro-ethernetip, ...)
+# contribute top-level subpackages under instro.* — without this, turning instro
+# into a regular package below would pin its __path__ to this directory alone
+# and hide those workspace trees. The exact spelling below (rather than
+# `from pkgutil import extend_path; ...`) matters: griffe/mkdocstrings pattern-match
+# this specific idiom to keep treating instro as a mergeable namespace for API docs.
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)
+
+from instro import cli, daq, dmm, eload, i2c, lib, modbus, psu, scope
+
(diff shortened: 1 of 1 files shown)


instro/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Typed Python API for test-and-measurement instruments."""
```
hoehner, line 1: Two quick things:

1. With the eager imports, import instro now loads a good number of dependencies up front (which I think comes with some minor timing considerations). Did you consider a module level __getattr__ that imports subpackages on first access, plus __dir__ returning the names so IPython tab completion still works? I think this would get the same UX with no time penalty. That said, the eager version is WAY simpler and more explicit, so if you considered this tradeoff already, we're happy to go with what you've got here.

2. A small regression test, if you're up for it since subpackage exposure is now a hand-maintained list. I could see someone adding a new category later could forget to register it here. A quick test asserting every subdirectory of instro/ appears in instro.__all__ and is importable would pin your fix in place. This can be a follow on ticket if you think it makes sense.

Thanks again, looking forward to landing this!

hoehner (changes requested):

instro/__init__.py
stefanv: How would you feel about adding a small dependency? We can easily handle this with, e.g., https://pypi.org/project/lazy-loader/

hoehner: Looks perfect - lets try it on for size.

PR #276, opened 2026-07-14
fix(ci): generalize maturin release publishing across mixed rust/python packages

cartercanedy:
## Summary

Fixes [INSTRO-469](https://linear.app/nominal-io/issue/INSTRO-469).

Two changes:

1. **Regenerate the stale `packages/instro-ethernetip/Cargo.lock`.** The `crates/instro-ethernetip` release bumped the crate `0.1.0 -> 1.0.0`, but that standalone package is excluded from the root cargo workspace so release-please's `cargo-workspace` plugin never touched its lockfile. The lock still pinned `instro-ethernetip 0.1.0`, so any `--locked` maturin build (wheel/sdist) would fail. Bumped the single entry to `1.0.0`; verified with `cargo check --locked --manifest-path packages/instro-ethernetip/Cargo.toml`.

2. **Generalize `release-please-publish.yml` off the hardcoded `instro-ethernetip` path** using an explicit `.instro-publish` marker file. Since this won't be the only mixed Rust/Python bridge package, the maturin build/publish flow now discovers its targets from a marker rather than directory sniffing:
   - A mixed Rust/Python package opts in by committing a `.instro-publish` file at its package root (added here for `packages/instro-ethernetip`).
   - New `discover-maturin` job walks `paths_released` and keeps each path with a `.instro-publish` marker (`[ -f "$path/.instro-publish" ]`), emitting a JSON matrix `[{path, name}]`.
   - `build-maturin-wheels` / `build-maturin-sdist` / `publish-maturin` fan out over that matrix (wheels also cross-product the existing OS/target platforms). Artifacts are namespaced per package with the package name last (`maturin-wheel-<os>-<target>-<name>`, `maturin-sdist-<name>`); each package downloads only its own via an exact-suffix pattern `maturin-*-<name>` (no trailing wildcard, so one package name being a prefix of another can't cross-contaminate uploads).
   - `publish-python` skips any marked package generically, so the plain-`uv build` path still owns pure-Python packages.

   The old EIP-only jobs (`build-eip-wheels`, `build-eip-sdist`, `publish-eip`) are removed. The sdist verifier `tests/ethernetip/check_ethernetip_sdist.py` moved to `tests/check_maturin_sdist.py` and is parameterized with `--package <path>` (resolving the maturin `python-source` root and workspace/path-dependency layout generically). Marker convention is documented in AGENTS.md and CONTRIBUTING.md.

   User-controlled/derived values used in `run:` blocks are passed through `env:` per the repo's workflow-injection convention.

## Type of change

- [x] Bug fix (`fix`)

## Verification

- `cargo check --locked --manifest-path packages/instro-ethernetip/Cargo.toml` — passes with the regenerated lock (failed before).
- Built the EIP sdist (`uv build --sdist --package instro-ethernetip`) and ran `python tests/check_maturin_sdist.py sdist --package packages/instro-ethernetip <sdist>` — verifies all expected wrapper manifests, Python sources, wrapper Rust sources, and the `crates/instro-ethernetip` path-dependency sources are present.
- Ran the `discover-maturin` marker logic against a simulated `paths_released` (`["packages/instro-ethernetip","packages/instro-daq-ni","crates/instro-ethernetip","."]`) → selects only `packages/instro-ethernetip`.
- `just check` (ruff format, mypy, ruff lint) — clean.
- YAML parses.

## Tests

- [x] Existing tests cover this change
- [ ] Unit tests added or updated — the sdist verifier is the regression coverage for the lock/packaging path; it's the same check generalized, now runnable per package.

## Checklist

- [x] PR title follows Conventional Commits
- [x] I have read CONTRIBUTING.md
- [x] Documentation updated if user-facing behavior changed (AGENTS.md + CONTRIBUTING.md marker convention; no runtime behavior)

## Notes for reviewers

- The full multi-platform wheel build + PyPI publish can only be exercised on the actual `push`-to-`main` release run; the individual maturin `build`/`sdist` steps and the discovery/verification logic are validated locally as above.
- Per-package trusted publishing must be configured on PyPI for each future maturin package (same requirement noted in AGENTS.md for the crate/PyPI publishers).
- New bridge package checklist: add the `packages/<name>/` maturin project (with its own committed `Cargo.lock`) and drop a `.instro-publish` marker — no workflow edit needed.

Link to Devin session: https://app.devin.ai/sessions/7ca354582bc843858d2b19512c5d8a7c
Requested by: @cartercanedy

### DIFF
--- a/.github/workflows/release-please-publish.yml
+++ b/.github/workflows/release-please-publish.yml
@@ -50,8 +50,7 @@ jobs:
           if [[ "$RELEASED_PATH" == crates/* ]]; then
             # Pure-Rust crates publish to crates.io, not PyPI.
             echo "skip=true" >> "$GITHUB_OUTPUT"
-          elif [ -f "$RELEASED_PATH/Cargo.toml" ] && [ -f "$RELEASED_PATH/pyproject.toml" ] \
-            && grep -q 'build-backend = "maturin"' "$RELEASED_PATH/pyproject.toml"; then
+          elif [ -f "$RELEASED_PATH/.instro-publish" ]; then
             # Mixed Rust/Python (maturin) packages are built and published by the maturin jobs below.
             echo "skip=true" >> "$GITHUB_OUTPUT"
           elif [ "$RELEASED_PATH" = "." ]; then
@@ -137,13 +136,7 @@ jobs:
           packages='[]'
           while IFS= read -r path; do
             [ -n "$path" ] || continue
-            case "$path" in
-              packages/*) ;;
-              *) continue ;;
-            esac
-            [ -f "$path/Cargo.toml" ] || continue
-            [ -f "$path/pyproject.toml" ] || continue
-            grep -q 'build-backend = "maturin"' "$path/pyproject.toml" || continue
+            [ -f "$path/.instro-publish" ] || continue
             name="$(basename "$path")"
             packages="$(jq -c --arg path "$path" --arg name "$name" '. + [{path: $path, name: $name}]' <<<"$packages")"
           done < <(jq -r '.[]' <<<"$PATHS_RELEASED")
--- a/.github/workflows/release-please-publish.yml
+++ b/.github/workflows/release-please-publish.yml
@@ -43,21 +43,42 @@ jobs:
           enable-cache: false
       - name: Resolve package name
         id: pkg
+        env:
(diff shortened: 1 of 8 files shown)


cartercanedy:
devin made a big stinky with this one

PR #280, opened 2026-07-15
fix(daq): standardize internal analog reads on lists

### LINKED ISSUE #278: chore: DAQDriverBase `_fetch_analog` should always return `list[Measurement]`
Currently the `_fetch_analog` method on the DAQDriverBase returns a single `Measurement` object if there is only one measurement to return and a `list[Measurement]` if there's more than one to return. To avoid the potential situation of trying to index into a single object or the need to write a guard around this, we should just have `_fetch_analog` always return a `list[Measurement]`.

sahilmathur254:
## Summary

* Make the internal `InstroDAQ._fetch_analog()` and `_software_timed_read()` helpers consistently return `list[Measurement]`.
* Preserve the existing public `read_analog()` contract by unwrapping a single measurement in the dispatcher.
* Add regression coverage for zero, one, and multiple measurements across hardware- and software-timed reads.

Fixes nominal-io/instro#278

## Type of change

- [X] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

* `pytest -q tests/daq`: 77 passed, 10 deselected.
* `ruff format --check instro/daq/daq.py tests/daq/test_daq_drivers.py`: passed.
* `ruff check instro/daq/daq.py tests/daq/test_daq_drivers.py`: passed.
* `git diff --check`: passed.

## Tests

- [X] Existing tests cover this change
- [ ] No tests — explain why:

## Checklist

- [X] PR title follows [Conventional Commits](<https://www.conventionalcommits.org/>) (e.g. `feat(driver): add support for Keysight E36300`)
- [X] I have read [CONTRIBUTING.md](<CONTRIBUTING.md>)

## Notes for reviewers

The internal helpers now provide the stable list return shape requested in nominal-io/instro#278, while callers of the public API continue to receive a single `Measurement` when exactly one timebase cluster is produced. The public documentation and examples are therefore unchanged.


maxleblang (changes requested):
Refer to comment added to the issue. We should also modify `_software_timed_read` and `read_analog` to make sure we get rid of `Measurement | list[Measurement]` function return signatures

cartercanedy:
@devin review

tests/daq/test_daq_drivers.py
```diff
@@ -847,6 +847,51 @@ def test_start_background_false_read_analog_fetches_from_buffer():
...
+    measurements = [
+        Measurement(channel_data={"ut.ai0": [float(index)]}, timestamps=[index]) for index in range(measurement_count)
+    ]
+    mock_driver._read_to_measurements.return_value = measurements
+
+    result = daq._fetch_analog()
+
+    assert result == measurements
+
+
```
cartercanedy: remove this completely redundant test

cartercanedy (changes requested):
Mechanically, it looks fine, but it's a breaking change to the api, going to sit on this

sahilmathur254:
Thanks @cartercanedy  I removed the redundant test in 581638e. The focused DAQ suite still passes with 77 tests.
Understood on the breaking API change and no rush on the review. I’ve kept the list-only behaviour aligned with the discussion on #278, and I’m happy to adjust once there is a decision to handle compatibility.

hoehner:
Thanks for the thorough work here, the tests and doc updates are great and the internal cleanup is exactly what the issue asked for.

One problem: we're at 1.4.0, and merging this as fix(daq)! would force a 2.0.0 release for this one method. We're not ready for a major version yet, so let's keep the cleanup without changing the public API:

- Keep _software_timed_read and _fetch_analog always returning list[Measurement], that part is great
- In read_analog(), keep the existing public behavior: move the measurements[0] if len(measurements) == 1 else measurements unwrap up into the dispatcher, and leave its signature as Measurement | list[Measurement].
- Retitle the PR to fix(daq) without the !, and revert the docs/examples/hardware-test changes since user-facing behavior no longer changes.
- One small doc fix to keep: the read_analog() -> Any line in daq.mdx should stay Any either way — that section documents the driver base, where read_analog() returns a raw device response.

When we're ready for 2.0 we'll do the full public switch, and most of this diff can be reused then so none of this work is wasted.

sahilmathur254:
Thanks @hoehner, makes sense. 
I’ve updated the implementation so `_software_timed_read()` and `_fetch_analog()` consistently return `list[Measurement]`, while `read_analog()`
retains the existing public `Measurement | list[Measurement]` behaviour and handles singleton unwrapping in the dispatcher.

I also reverted the public documentation, examples, and hardware test changes.
Regression coverage now exercises zero, one, and multiple measurements across both software and hardware paths.

maxleblang (dismissed):
This looks good to me. Only other nit I have is we could consolidate the two tests into one single `test_read_analog_preserves_public_return_shape` test that tests both hw timed and sw timed.

sahilmathur254:
Thanks. I’ll consolidate the hw and sw timed cases into a single parameterised `test_read_analog_preserves_public_return_shape` test, keeping both dispatch paths covered, and then rerun the focused DAQ suite and Ruff checks. No behaviour changes are needed.

maxleblang (dismissed):
Test LGTM

sahilmathur254:
Hi @cartercanedy , the latest revision preserves the public read_analog return shape and only standardises the internal helpers. Both paths are covered, and Max has approved. Could you please re-review and update the outstanding change request?

cartercanedy:
@greptileai

cartercanedy (approved):

cartercanedy:
Thanks, @sahilmathur254!

PR #284, opened 2026-07-16
fix(nidaqmx): validate digital port width

### LINKED ISSUE #245: NIDAQDriver: DigitalPortWidth is not validated against the physical port width
**Versions:** instro 1.1.0, instro-daq-ni 1.0.0, Python 3.14, Windows 11

## Bug

`configure_di_port_channel` / `configure_do_port_channel` accept any `DigitalPortWidth` without checking the physical port's actual line count. Two consequences:

1. **Reads return values the hardware cannot produce.** The active-LOW inversion XORs the *declared* width's mask, so an 8-line port declared `WIDTH_16` reads `0xFFFF` — the upper 8 bits are pure artifacts of the mask, not data.
2. **Writes are silently truncated.** `write_digital_port` bounds-checks against the declared width only; a value that fits 16 bits but exceeds the port's 8 lines is accepted and DAQmx masks the extra bits with no error.

## Repro

Simulated PCIe-6363 as `Dev1` (`Dev1/port1` has 8 lines):

```python
from instro.daq import InstroDAQ
from instro.daq.drivers.ni import NIDAQDriver
from instro.daq.types import DigitalPortWidth, Direction, Logic

daq = InstroDAQ(name="daq", driver=NIDAQDriver(device_id="Dev1"))
with daq:
    daq.configure_digital_port(Direction.INPUT, "Dev1/port1", Logic.LOW, DigitalPortWidth.WIDTH_16, alias="bus")
    print(daq.read_digital_port("bus").channel_data)
    # -> {'daq.bus': [65535.0]}  # 0xFFFF from an 8-line port (raw read was 0, XORed with the 16-bit mask)

    daq.configure_digital_port(Direction.OUTPUT, "Dev1/port1", Logic.HIGH, DigitalPortWidth.WIDTH_16, alias="obus")
    daq.write_digital_port("obus", 300)  # exceeds the 8 physical lines; accepted, silently masked
```

## Workaround

Always declare the `DigitalPortWidth` that matches the physical port (e.g. `WIDTH_8` for `Dev1/port1`); the mask math and bounds checks are then correct.

## Suggested fix

At configure time, query the port's real width from DAQmx (`nidaqmx.system.PhysicalChannel.di_port_width` / `.do_port_width`) and either reject a mismatched `DigitalPortWidth` with a `ValueError`, or drop the parameter for NI entirely and derive the width from the hardware.

maxleblang:
## Summary

Fetches the port width of a digital port and compares it to the passed in port width. If there's a mismatch a `ValueError` is raised.

## Type of change

- [x] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Running `tests/daq/ni/test_nidaq_port.py` results in 4 passes. Additionally, the full test suite passes as expected.

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers


dalvarez-204 (approved):
Looks good, nice fix for the NI driver like the MCC driver does. Test coverage looks solid too for DI and DO.

PR #295, opened 2026-07-21
feat(transports): add composable ModbusDriver transport

wharwood:
## Summary

Makes Modbus a first-class, composable transport in `instro/lib/transports/`, paralleling `VisaDriver`. A new `ModbusDriver` owns the Modbus wire layer so register-mapped instrument drivers can compose it the same way SCPI drivers compose `VisaDriver`. The existing `ModbusDevice` HAL is refactored to compose it rather than reimplement it.

Closes INSTRO-492. Prerequisite for the process-controller category (INSTRO-487 — Watlow, Red Lion, etc.).

## Type of change

- [ ] Bug fix (`fix`)
- [X] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Adds tests.

## Tests

- [ ] No tests — explain why:

## Checklist

### DIFF
--- a/instro/lib/transports/modbus.py
+++ b/instro/lib/transports/modbus.py
@@ -1,0 +1,395 @@
+"""Modbus transport driver. Wraps pymodbus; callers own register maps, this owns I/O, framing, and locking."""
+
+from __future__ import annotations
+
+import functools
+import struct
+import threading
+from typing import TYPE_CHECKING, Literal
+
+from pydantic import BaseModel, Field
+from pymodbus.exceptions import ConnectionException as PymodbusConnectionException
+
+if TYPE_CHECKING:
+    from pymodbus.client import ModbusSerialClient, ModbusTcpClient
+
+
+class TCPConnection(BaseModel):
+    """Modbus TCP connection configuration."""
+
+    transport: Literal["tcp"] = "tcp"
+    host: str
+    port: int = Field(default=502, ge=1, le=65535)
+    unit_id: int = Field(default=1, ge=0, le=255)
+    timeout: float = Field(default=3.0, gt=0, description="Response timeout in seconds")
+
+
+class RTUConnection(BaseModel):
+    """Modbus RTU (serial) connection configuration."""
+
+    transport: Literal["rtu"] = "rtu"
+    port: str  # e.g., "/dev/ttyUSB0" (Linux), "/dev/cu.usbserial-1234" (macOS), "COM3" (Windows)
+    baudrate: int = 9600
+    parity: Literal["N", "E", "O"] = "N"
+    stopbits: Literal[1, 2] = 1
+    bytesize: Literal[5, 6, 7, 8] = 8
+    unit_id: int = Field(default=1, ge=0, le=255)
+    timeout: float = Field(default=3.0, gt=0, description="Response timeout in seconds")
+
+
+ConnectionType = TCPConnection | RTUConnection
+
+# Modbus protocol vocabulary. Single source of truth; ``instro.modbus.types`` re-exports these.
+RegisterType = Literal["holding", "input", "coil", "discrete"]
+DataType = Literal["uint16", "int16", "uint32", "int32", "uint64", "int64", "float32", "float64", "bool"]
+
+
+def _modbus_op(fn):
+    """Acquire the lock, run the operation, and clear dead sockets on failure.
+
+    The pymodbus sync client does not auto-reconnect between operations
+    (``reconnect_delay`` is async-only). It does call ``connect()`` before every
+    operation, creating a new socket when ``self._client.socket is None``. So on
+    a transport error we close the dead socket and re-raise — the next call gets
+    a fresh connection.
+    """
+
+    @functools.wraps(fn)
+    def wrapper(self, *args, **kwargs):
+        with self._lock:
+            if self._client is None:
+                raise RuntimeError("Modbus client not connected. Call open() first.")
+            try:
+                return fn(self, *args, **kwargs)
+            except (OSError, ConnectionError, PymodbusConnectionException):
+                self._client.close()
+                raise
+
+    return wrapper
+
+
+class ModbusDriver:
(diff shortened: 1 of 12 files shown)


instro/lib/transports/modbus.py
```diff
@@ -0,0 +1,395 @@
...
+        long_swap: bool = False,
+    ) -> int | float:
+        """Read ``address`` as ``data_type``, dispatching by ``register_type`` and decoding across registers."""
+        match register_type:
+            case "holding":
+                raw_regs = self.read_holding_registers(address, self.register_count(data_type))
+            case "input":
+                raw_regs = self.read_input_registers(address, self.register_count(data_type))
+            case "coil":
+                return int(self.read_coils(address, 1)[0])
```
maxleblang: For type `coil` and `discrete` we hardcode using `bool` values, which is correct. However, we never check the input `data_type` on these calls so a user can read a `coil` with `data_type` `int16` and the call will be successful. We should at least raise a warning here (and in `write_typed`) that `coil` and `discrete` only work on `bool` data.

instro/lib/transports/modbus.py
```diff
@@ -0,0 +1,395 @@
...
+        """Encode ``value`` as ``data_type`` and write it to ``address``, dispatching by ``register_type``."""
+        match register_type:
+            case "holding":
+                encoded = self.encode_value(value, data_type, byte_swap, word_swap, long_swap)
+                if len(encoded) == 1:
+                    self.write_holding_register(address, encoded[0])
+                else:
+                    self.write_holding_registers(address, encoded)
+            case "coil":
+                self.write_coil(address, bool(value))
```
maxleblang: Similarly here we should raise a warning if the value they passed in isn't a 1 or 0 ("Casted <VALUE> as a bool for coil write" or something like that)

instro/lib/transports/modbus.py
```diff
@@ -0,0 +1,395 @@
...
+        word_swap: bool = False,
+        long_swap: bool = False,
+    ) -> int | float:
+        """Read ``address`` as ``data_type``, dispatching by ``register_type`` and decoding across registers."""
+        match register_type:
+            case "holding":
+                raw_regs = self.read_holding_registers(address, self.register_count(data_type))
+            case "input":
+                raw_regs = self.read_input_registers(address, self.register_count(data_type))
+            case "coil":
```
hoehner, line 242: Could we (or should we) enforce that the "data_type" passed in here for coils and discrete MUST be a boolean type? I think what ever we decide we should do on the write side as well for coils. Not sure what this breaks but I like the certainty of never having to worry about numeric coercion to a boolean (not sure if the ModbusDevice does it this way but I'm fine if we need to diverge here to get this right from the start).

instro/lib/transports/modbus.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def __init__(self, connection: TCPConnection | RTUConnection) -> None:
+        """Construct from a ``TCPConnection`` or ``RTUConnection``."""
+        self._connection = connection
+        self._client: ModbusTcpClient | ModbusSerialClient | None = None
+        self._lock = threading.RLock()
+
+    @property
+    def is_open(self) -> bool:
+        """Whether the underlying Modbus client is currently connected."""
```
hoehner: Let's soften this up to say "has been opened". I think this returns True all the time after we try to open the connection (even if we disconnect). I can't quite recall if that is intentional for the reconnection logic but it would be at least worth making this language more accurate to how this check behaves.

hoehner (dismissed):
2 minor nits!

tests/lib/test_modbus_driver.py
```diff
@@ -0,0 +1,332 @@
...
+        assert ModbusDriver.decode_registers(encoded, data_type) == value
+
+    @pytest.mark.parametrize("data_type,rel", [("float32", 1e-6), ("float64", 1e-12)])
+    def test_float_roundtrip(self, data_type, rel):
+        encoded = ModbusDriver.encode_value(3.14159265, data_type)
+        assert ModbusDriver.decode_registers(encoded, data_type) == pytest.approx(3.14159265, rel=rel)
+
+    @pytest.mark.parametrize("word_swap", [False, True])
+    @pytest.mark.parametrize("byte_swap", [False, True])
+    def test_swap_roundtrip(self, word_swap, byte_swap):
```
maxleblang, line 342: Why do we not test `long_swap`?

tests/lib/test_modbus_driver.py
```diff
@@ -0,0 +1,332 @@
...
+    drv = ModbusDriver(TCPConnection(host="127.0.0.1", port=TEST_PORT))
+    drv.open()
+    yield drv
+    drv.close()
+
+
+# ============ Lifecycle ============
+
+
+class TestLifecycle:
```
maxleblang, line 122: Can we add a test that verifies we can reconnect after a transport error has been raised? Something like

maxleblang (commented):
Just some nits. Overall looks good

instro/lib/transports/modbus.py
wharwood: I think @hoehner was opting for a strict reject rather than a warn here. I tend to agree with him. Are you good with that direction?

tests/lib/test_modbus_driver.py
wharwood: Good catch.

tests/lib/test_modbus_driver.py
wharwood: Yes.

instro/lib/transports/modbus.py
wharwood: Done. The docstring is now explicit on what the behavior is.

instro/lib/transports/modbus.py
wharwood: Yes, I think this is the right choice. I'll look into this.

maxleblang (approved):
LGTM

hoehner (approved):

Issue #223, opened 2026-07-06
refactor(daq): restructure the InstroDAQ/DAQDriverBase interface


wharwood:
The shared DAQ interface (InstroDAQ and DAQDriverBase) has accreted inconsistently as drivers were added, and it now blocks common acquisition setups while pushing internal concepts onto users. It assumes analog channels are voltage, so a user who plugs a thermocouple module (NI 9213) or a current-input module (NI 9203) into a cDAQ chassis cannot configure it, even though reading temperature or a 4-20 mA loop is what that hardware does; the channel model has nowhere to carry the parameters those modes need, such as thermocouple type, cold-junction handling, and current range. The channel model also takes a different shape for analog and digital channels, and it tangles three orthogonal concerns (signal mode, direction, and wiring) into one representation, so equivalent concepts appear in non-parallel forms across the type hierarchy. Configuring a channel requires importing enums and internal types (direction, logic, terminal config), which leaks HAL and vendor internals into user code for what should be string-level configuration. The shared surface also carries concepts of uneven value: relay and mux routing specific to one vendor (the Keysight 34980A), a digital port-versus-line split that is really a driver-internal detail, and HAL-side scaling that duplicates conditioning the device already performs. The interface spreads read and write across narrow per-type methods (read_analog, read_digital_line, read_digital_port, write_analog_value, write_digital_line) backed by a large set of overlapping configure_* methods on the driver contract, and the line between what the HAL owns and what the driver owns stays blurry. The result is a DAQ layer that resists real instrumentation setups, new signal types, and the maintainer reading a driver; fixing it means reworking the base interface, not patching individual drivers.

maxleblang:
Closed by #426

PR #303, opened 2026-07-24
feat(instrodaq): unified read and write functions

maxleblang:
## Summary

This PR adds unified `read` and `write` functions to `InstroDAQ` that route to either `read_analog/digital` or `write_analog/digital` depending on the type of channel it's operating on. This cleans up the surface a ton and makes reading and writing data with an `InstroDAQ` much easier.

Closes INSTRO-501

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

ver 501

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

The new `read()` function wraps the existing `read_analog()` function that still raises an error when trying to read samples when the background daemon is running. We need to fix this functionality.


instro/daq/daq.py
```diff
@@ -902,6 +903,77 @@ def _scale_analog_measurement(self, measurements: list[Measurement]) -> list[Mea
...
+            if channel in ao:
+                commands.append(self.write_analog_value(channel, value, **kwargs))
+                continue
+            if channel in do:
+                # NOTE: Planning on ripping out port support
+                if isinstance(do[channel], DigitalPortChannel):
+                    commands.append(self.write_digital_port(channel, int(value), **kwargs))
+                    continue
+                # Why does this function accept the data as an int? Shouldn't it only accept a bool?
+                commands.append(self.write_digital_line(channel, int(value), **kwargs))
```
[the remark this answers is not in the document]
maxleblang: This is existing behavior and fixing is out of scope of this ticket

instro/daq/daq.py
```diff
@@ -984,6 +985,77 @@ def _scale_analog_measurement(self, measurements: list[Measurement]) -> list[Mea
                     measurement.channel_data[f"{self.name}.{ch_name}"] = scaled_values
         return measurements
 
+    def read(
```
maxleblang: I now have our `read` returning a dict instead of just a list of measurements. To me this is much more usable than having the user find their measurements in a list. Was there a reason we were just returning a list of measurements from `read_analog()`?

maxleblang:
Work for #223

hoehner:
Just noting one ergonomics thing - it seems weird to me to have to do a .read(ALIAS)[ALIAS] to get the measurement. If we have just a single thing we want to read, we should just return the single measurement, not require the user to do another thing with the same alias name. 

In the case where we are reading more than one alias, I think the logic still holds. We should just return a list of measurements that are pulled from the dict in the read method. We don't necessarily need to return ALL of the items from the DAQ, but we should make the ergonomics of the API much simpler by not requiring the dict lookup after a .read() method is called. I think this would actually make the examples a lot cleaner overall since you're not having to specify the ALIAS twice to get a measurement object or list of measurement objects out.

instro/daq/daq.py
```diff
@@ -1085,6 +1086,78 @@ def _scale_analog_measurement(self, measurements: list[Measurement]) -> list[Mea
...
+        else:
+            aliases = [channels] if isinstance(channels, str) else list(channels)
+            # Validate up front so a bad alias can't leave earlier channels already read from hardware.
+            if unknown := [a for a in aliases if a not in ai and a not in di]:
+                raise KeyError(f"Input channel(s) {unknown} not configured. Configured input channels: {[*ai, *di]}.")
+
+        # Read every analog channel once, then map each channel key to the Measurement it came back in.
+        analog_source: dict[str, Measurement] = {}
+        if any(alias in ai for alias in aliases):
+            analog = self.read_analog(**kwargs)
```
hoehner: Am I thinking about this correctly that we would only use the read() method when we're not ALSO using the background daemon? If the answer is no, then doing a read_analog() while the background daemon is still running would mean we're essentially throwing away samples for the aliases not listed in this single read() method. Let me know if I'm not understanding this correctly.

hoehner:
We should reject aliases that are the same for different channel types - we shouldn't have an ai channel and a di channel using the same alias (same goes for ao and do). All aliases on a single InstroDAQ should be unique.

hoehner:
We should probably only allow 1s and 0s for aliases that map to digital channels. Additionally when writing a port we shouldn't allow anything other than integer numbers (no floats allowed for port aliases).

instro/daq/daq.py
maxleblang: Yes that's correct. As it stands right now, `read()` is meant to only unify `read_analog()` and `read_digital()` via some routing logic. It isn't changing any underlying functionality though. Thinking about it more now, it might make sense to also handle the background daemon case as well and add routing logic to `get_channel()` so that `read()` is totally universal. What're your thoughts here?

maxleblang:
> We should reject aliases that are the same for different channel types - we shouldn't have an ai channel and a di channel using the same alias (same goes for ao and do). All aliases on a single InstroDAQ should be unique.

We already handle this in `InstroDAQ` with `_reject_duplicate_channel()` which checks across all configured channel aliases for a given `InstroDAQ`'s driver.

hoehner:
@claude review this PR

maxleblang:
Added functionality to `write_batch()` so that a user can specify `continue_on_failed_write`. This will allow a user to try every command in a `write_batch()` call no matter what, or stop and raise on a failed command. We also have verbose logs (at the debug level) to tell the user what commands are succeeding and what are failing.

Also added an example that constantly streams values out to a cDAQ AO card so that we can manually play with and verify this behavior:
no_continue
^This is with `continue_on_failed_write` set to `False`. I unplugged the cDAQ while streaming and it failed on the write command and raised.

yes_continue
^This is with `continue_on_failed_write` set to `True. I unplugged the cDAQ while streaming and it continued to try to write to the outputs. This was in a loop so once I plugged the cDAQ back in it started successfully writing again.

Also added hardware tests to validate the `continue_on_failed_write` functionality. Also added hardware tests to test the following situations for `read()/read_batch()` and `write()/write_batch()`:
- Normal commanding of a single channel
- Normal commanding of multiple channels to multiple values
- Commanding of multiple channels where one of the channels isn't valid or defined
- Reading of a single channel
- Reading of multiple channels
- Reading of multiple channels with a background daemon
- Reading of invalid single channel
- Reading of multiple channels where one is invalid

Validated these tests pass on all supported hardware (except KeySight @hoehner)
t8
mcc
ni
t4
t7

hoehner (approved):

PR #307, opened 2026-07-27
feat(awg): add Rigol-DG1022Z driver

### LINKED ISSUE #158: driver: adding driver for Rigol DG1022 (AWG)

LizzyC-115:
## Summary

Adds `RigolDG1022Z`, a SCPI/VISA driver for the Rigol DG1022Z two-channel arbitrary waveform generator (DG1000Z command family), implementing `AWGDriverBase`: Sine, Square, Sawtooth, Triangle, Pulse, Arbitrary, and StaticValue (DC) waveforms, plus amplitude/offset, output enable, output load, and channel phase alignment. Includes mocked software tests and a hardware smoke-test suite. Registers `RigolDG1022Z` in `instro/unstable/awg/drivers/__init__.py` and adds the AWG category to the README's "Supported devices" table.

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification
Rigol driver hardware test passing

Rigol driver software tests passing

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

**Why per-point downloads instead of bulk:**

ARB waveform downloads intentionally use per-point `:TRAC:DATA:VALue` writes rather than a single bulk `:TRAC:DATA VOLATILE` write. This works around a firmware bug on the bench unit where bulk block downloads — including Rigol's own manual example, sent verbatim — hang the instrument's firmware over USB and require a power cycle to recover. With the per-point approach, both USB and Ethernet connections reliably generate arbitrary waveforms.

The tradeoff is speed: bulk downloads are roughly two orders of magnitude faster on transports where they don't crash the firmware, so per-point writes add meaningful latency for larger waveforms. See the speed comparison below.

Later changes to this driver will include choosing bulk download where preferable instead of per-point. In other words:

LAN + ~32KB or less download size -> Bulk download
LAN + ~33KB or more download size -> Per point download
USB + any size download -> Per point download

Screenshot 2026-07-27 at 3 58 21 PM


LizzyC-115:
Leaving burst, modulation, sweep, and harmonics to a later PR/Milestone

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -0,0 +1,185 @@
...
+    def get_amplitude(self, channel: int) -> tuple[float, AmplitudeMeasurementUnit]:
+        _check_channel(channel)
+        with self._visa.lock():
+            unit = AmplitudeMeasurementUnit(self._visa.query(f":SOUR{channel}:VOLT:UNIT?").strip())
+            amplitude = float(self._visa.query(f":SOUR{channel}:VOLT?"))
+        return amplitude, unit
+
+    def set_offset(self, channel: int, offset: float) -> None:
+        _check_channel(channel)
+        self._visa.write(f":SOUR{channel}:VOLT:OFFS {offset}")
```
nhschwab, line 148: Based on the software test, it looks like the SCPI command to get the offset in DC mode should be `APPL?`. Can you double check, and if true add a conditional to send the correct command?

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -0,0 +1,185 @@
...
+                        f" per download, got {num_points}"
+                    )
+                self._visa.write(f":SOUR{channel}:APPL:ARB {waveform.sample_rate_hz}")
+                self.check_errors()
+                self._visa.write(f":SOUR{channel}:TRAC:DATA:POIN VOLATILE,{num_points}")
+                self.check_errors()
+                for point, sample in enumerate(waveform.samples, start=1):
+                    decimal_value = round((sample + 1) / 2 * 16383)
+                    self._visa.write(f":SOUR{channel}:TRAC:DATA:VAL VOLATILE,{point},{decimal_value}")
+                    self.check_errors()
```
nhschwab, line 92: Checking the error queue after every single sample is very expensive. Might be worth checking errors at the end of the loop

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
LizzyC-115: Added a check for DC to the logic

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
LizzyC-115: Moved check_errors outside the loop

docs/guides/instrumentation/quickstart.mdx
```diff
@@ -81,7 +81,7 @@ That's the whole loop: construct → `open()` → configure → measure → `clo
...
+You can drive the instrument with explicit `open()` / `close()` calls, or let Python do it for you with a `with` block. Both are first-class. Pick whichever fits your script. Every `instro` instrument is a context manager. `__enter__` calls `self.open()`, `__exit__` calls `self.close()` (including when an exception escapes the block, so the daemon and publishers always get torn down). This works for every built-in category (`InstroPSU`, `InstroDMM`, `InstroDAQ`, `InstroELoad`, `InstroAWG`, `I2CInterface`, …) and for any custom `Instrument` subclass (see [Custom instruments](/instrumentation/custom-instruments)).
```
nhschwab: Why are we replacing `InstroFlowController` instead of just adding `InstroAWG` to the list?

docs/reference/src/index.md
```diff
@@ -16,8 +16,8 @@ The SDK is organized into several key components:
 
 - **Instruments**: High-level, vendor-agnostic interfaces for common instrument types:
   [DAQ](instruments/daq.md), [DMM](instruments/dmm.md), [PSU](instruments/psu.md),
-  [Electronic Load](instruments/eload.md), [Flow Controller](instruments/flowcontroller.md),
-  [Scope](instruments/scope.md), and [I2C](instruments/i2c.md).
+  [Electronic Load](instruments/eload.md), [Scope](instruments/scope.md),
+  [I2C](instruments/i2c.md), and [AWG](instruments/awg.md).
```
nhschwab: Same here, why deleting flow controller?

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -0,0 +1,191 @@
...
+    def check_errors(self) -> None:
+        """Drain ``:SYSTem:ERRor?`` and raise on the first non-zero code."""
+        while True:
+            resp = self._visa.query(":SYST:ERR?")
+            parts = resp.split(",", 1)
+            code = int(parts[0])
+            if code == 0:
+                return
+            msg = parts[1].strip().strip('"') if len(parts) > 1 else "Unknown error"
+            raise RuntimeError(f"Rigol DG1022Z reported error {code}: {msg}")
```
nhschwab: This method isn't actually draining the device's error queue, it is either okay because error code is 0 or it is raising on the oldest error in the queue. As a result, the loop is dead code. Looks like this was pattern matched against the existing check_errors method in the keysight 1200 scope driver, which is also wrong.

For this driver specifically, we actually do care about draining the error queue because we do send batch commands in the arbitrary waveform download command. I propose we simplify this and make the error check similar to most all other instrument drivers in the codebase by just checking if the error code is 0 and otherwise raising. Then we should move the error check back into the point-by-point loop in the download method. Let's include an in-line comment in that method to explain the decision and document the performance tradeoff.

nhschwab: We should also cut a new issue to go back and fix the error check method in the keysight scope driver

LizzyC-115: Fixed. Simplified the check errors following the PSU formatting and added a comment to explain in set_waveform. Issue for future iterations here: #332

nhschwab (approved):

PR #308, opened 2026-07-27
feat(psu): add get_voltage_setpoint() and get_current_setpoint() to PSUDriverBase

dalvarez-204:
## Summary
- Adds two new optional `PSUDriverBase` methods, `get_voltage_setpoint()` and `get_current_setpoint()`, plus matching `InstroPSU` wrapper methods — same pattern as the existing OVP/OCP getters (default `NotImplementedError`; drivers opt in).
- These read back the *configured* voltage setpoint and current setpoint (typically a plain `VOLT?`/`CURR?` SCPI query), distinct from `get_voltage()`/`get_current()` which measure actual output (`MEAS:VOLT?`/`MEAS:CURR?`).
- No concrete driver implements them yet — each vendor driver needs its own override before this is usable against real hardware. Verified the general pattern against BK914X's existing `get_overcurrent_protection_level` shape but didn't have hardware on hand to validate a concrete implementation.

## Test plan
- [x] `just check` (format, mypy, lint) passes
- [x] `tests/psu/test_psu_drivers.py` — new parametrized `NotImplementedError` tests for both methods, plus `InstroPSU`-level delegation tests (33/33 passing)
- [x] Full `tests/psu/` suite passes (492 tests)
- [x] Docs updated: Published channels table, Method Reference table, and the `PSUDriverBase` interface listing in `docs/guides/instrumentation/psu.mdx`
- [ ] Hardware validation against a concrete driver implementation — deferred until hardware is available

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/guides/instrumentation/psu.mdx
+++ b/docs/guides/instrumentation/psu.mdx
@@ -226,6 +226,8 @@ Every measurement/command call produces a channel keyed under `{name}.{descripto
 | `get_current(channel=N)` | `ch{N}.current` | telemetry |
 | `output_enable(channel=N)` | `ch{N}.enabled.cmd` | command |
 | `get_output_status(channel=N)` | `ch{N}.enabled` | telemetry |
+| `get_voltage_setpoint(channel=N)` | `ch{N}.voltage.setpoint` | telemetry |
+| `get_current_limit(channel=N)` | `ch{N}.current.limit` | telemetry |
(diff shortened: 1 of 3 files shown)


hoehner:
Lets just make the names similar - get_voltage_setpoint and get_current_setpoint. This gets the idea across for both.

hoehner (changes requested):
Minor request to make names similar because they're querying similar things (setpoints set by the user).

hoehner (approved):

Issue #305, opened 2026-07-26
fix(daq): reject AI start without hardware timing


cartercanedy:
# AI issue report, currently treating as unverified

## Problem

`InstroDAQ.start()` can enter an invalid state when analog-input channels exist but `configure_ai_sample_rate()` has not been called.

The HAL currently:

1. Calls `driver.start()` without validating `ai_hw_timing_config`.
2. Registers `_fetch_analog()` whenever AI channels exist.
3. Starts the background thread.

Driver behavior is inconsistent. Some drivers reject the missing timing configuration, but `Keysight34980A.start()` sends `INIT` and succeeds. Its background thread then repeatedly calls `_fetch_analog()`, which raises because no hardware timing is configured. The base daemon catches that exception and remains alive. Meanwhile, `read_analog()` sees no timing configuration and dispatches a software-timed read despite the live, broken daemon.

This invariant violation predates nominal-io/instro#278. Moving the daemon check to the top of `read_analog()` would only mask one symptom; the invalid acquisition should never start.

## Expected behavior

Starting an AI acquisition without AI hardware timing should fail before any driver acquisition or background thread starts. Software-timed reads should continue to work without calling `start()`.

## Acceptance criteria

* Validate the AI timing prerequisite at the `InstroDAQ` boundary before calling `driver.start()`.
* Raise a clear error when AI acquisition is started without `configure_ai_sample_rate()`.
* Do not leave the driver partially started or create a background thread on failure.
* Preserve `start(background=False)` for manually fetched, hardware-timed AI acquisition.
* Preserve valid non-AI start behavior where applicable.
* Add regression coverage for the rejected state and normal software/hardware-timed paths.

## Related work

* nominal-io/instro#61 changed daemon ownership to use thread liveness but did not enforce this prerequisite.
* nominal-io/instro#73 proposes broader timing-domain support; this issue covers the existing AI-only invariant independently.
* nominal-io/instro#223 is a broader DAQ interface redesign and does not explicitly track this failure mode.

hoehner:
Software timed acquisitions that use a background daemon should still be allowed to call start(). The background daemon serves two purposes, handing the incoming points for a hardware timed acquisition AND running the daemon to do software timed queries for software timed acquisition. Not sure how that changes this ticket, but some of this described behavior is intended (could be cleaned up or more obvious).

maxleblang:
[...]

PR #325, opened 2026-07-29
feat(instrodaq): add support for sw timed background daemon

maxleblang:
## Summary

This PR adds support for running a background daemon that is running software timed acquisition. The user can now configure a software timing sample rate and have a background daemon acquire samples from the DAQ at the defined rate (the work time of the daq read is the ceiling for sw sample rate).

We also add new examples highlighting this new functionality. Additionally, for DAQs that support multiple `InstroDAQ` configurations (NIDAQ) we support configuring a hw timed and a sw timed background daemon that can run in parallel.

Here are the 6 scenarios a user can now configure and start an `InstroDAQ` instance, and what they result in:

1. User configures hw timing and calls `start()` -> DAQ hw timed acquisition started and background daemon started
2. User configures hw timing and calls `start(background=False)` -> DAQ hw timed acquisition started
3. User configures sw timing and calls `start()` -> background daemon started at specified sample rate
4. User configures sw timing and calls `start(background=False)` -> ERROR RAISED: sw timed continuous acquisition can't be started with no background daemon
5. User doesn't configure any timing and calls `start()` -> background daemon started at default (1 Hz) sample rate
6. User doesn't configure any timing and calls `start(background=False)` -> ERROR RAISED: sw timed continuous acquisition can't be started with no background daemon

Closes INSTRO-524

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

New examples behaved as expected. Test suite passes and new NIDAQ hardware test passes as well.

ver

ver hardware

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

This has only been tested on an NI cDAQ and a LabJack T4. I also need test validation on a MCC and Keysight just to ensure the background software acquisition works as expected. This just entails running the `daq_read_analog_sw_timed.py` example for both these DAQs and ensuring that data is being read properly via software acquisition in the background.


examples/daq/daq_read_analog_sw_timed_no_background.py
maxleblang, line 1: This is the previous `daq_read_analog_sw_timed.py` examples

instro/daq/daq.py
```diff
@@ -609,13 +690,30 @@ def _software_timed_read(self, **kwargs) -> list[Measurement]:
...
         return self._scale_analog_measurement(measurements)
 
+    def _fetch_analog_sw_timed(self, **kwargs) -> list[Measurement]:
+        """Daemon-cycle software-timed read; warns when the read can't sustain the configured period."""
+        read_start = time.perf_counter()
+        measurements = self._software_timed_read(**kwargs)
+        read_time = time.perf_counter() - read_start
+
+        period = self._background_config.interval
+        if self.is_sw_timing_configured and read_time > period:
```
maxleblang: Raise a warning if the loop work time is longer than requested background interval

instro/daq/daq.py
```diff
@@ -534,16 +569,59 @@ def configure_ai_sample_rate(
...
+        **kwargs,
+    ):
+        """Configure the software-timed polling rate for AI channels.
+
+        Args:
+            sample_rate: Rate (Hz) at which the background daemon polls AI channels.
+        """
+        self._require_open()
+        if self.is_hw_timing_configured:
+            raise HWTimingException(
```
nhschwab: Making this a HWTimingException feels like it will confuse users. Maybe we created a new typed exception TimingConfigError that is a subclass of InstroError. There are a few other spots where we shouldn't raise HWTimingException as well

instro/daq/daq.py
nhschwab: This warning will spam since the embedded 1/read_time will change each time. Also, warnings.warn feels like the wrong channel, we should probably use logger instead.

instro/daq/daq.py
```diff
@@ -429,15 +430,25 @@ def di_hw_timing_config(self) -> HWTimingConfig | None:
...
-    # Need to ensure background interval never adds a wait for InstroDAQ
+    @property
+    def is_hw_timing_configured(self) -> bool:
+        """True once ``configure_ai_hw_sample_rate()`` has programmed the AI sample clock."""
+        return self.ai_hw_timing_config is not None
+
+    @property
+    def is_sw_timing_configured(self) -> bool:
+        """True once ``configure_ai_sw_sample_rate()`` has set a software-timed polling rate."""
+        return self._background_config.interval > 0
```
nhschwab: Make sure that all dispatchers encode the same precedence rules for consistency, probably hw timed first.

The sw timed check can be bypassed with `InstroDAQ(..., background_config=BackgroundConfig(interval=0.5))`, so we need to make sure that it is documented that this results in sw timed. The user can no longer use the background daemon to pace the hw timed acqusition.

nhschwab: Actually I think the second part is false, but worth double checking

instro/daq/daq.py
```diff
@@ -534,16 +569,59 @@ def configure_ai_sample_rate(
...
+            self._define_background_daemon()
+            super().start()
+            return
+
+        if not self.is_hw_timing_configured:
+            raise HWTimingException(
+                f"Cannot start DAQ '{self.name}' without AI timing configured. "
+                "Call configure_ai_hw_sample_rate() or configure_ai_sw_sample_rate() first."
+            )
+
```
nhschwab: This is a breaking change. Even though it's an improvement, what previously would be silent now loudly fails

instro/daq/daq.py
```diff
@@ -578,17 +659,17 @@ def read_analog(
...
         self._require_open()
-        if self.ai_hw_timing_config:
-            if self._background_thread and self._background_thread.is_alive():
-                # Background daemon running. The user can't pull from the buffer mid-flight.
-                # TODO revisit with INSTRO-149 issue ticket.
-                raise RuntimeError("Cannot read analog data while background acquisition daemon is running")
+        if self._background_thread and self._background_thread.is_alive():
+            # Background daemon running. The user can't pull from the buffer mid-flight.
+            # TODO revisit with INSTRO-149 issue ticket.
+            raise RuntimeError("Cannot read analog data while background acquisition daemon is running")
```
nhschwab, line 1029: This is a breaking change, a sw-timed read_analog with a background daemon now raises

Cargo.lock
nhschwab, line 1: Don't need this version bump

uv.lock
nhschwab, line 1: Don't need these bumps

instro/daq/daq.py
maxleblang: I didn't think `logger.warn` dedupes the same way that `warnings.warn` does. Also I found that rounding `1/read_time` to one decimal place usually doesn't change too much so the warning doesn't spam.

Do we want to add logic to just warn once using `logger.warn`?

instro/daq/daq.py
maxleblang: Makes sense. I'll add and change here and we can work on converting those other spots in another PR.

instro/daq/daq.py
maxleblang: @hoehner What should we do here? This is an existing bug called out by issue #305 but fixing it is a breaking change. The break is raising a correct error now though.

instro/daq/daq.py
maxleblang: This is an odd one because there is was a world where previously the user could not configure an `ai_hw_timing_config`, called `start()` (which would essentially do nothing), and then call `read_analog` to do a software read. This would break that behavior. However, that behavior is wrong and the background daemon would constantly be logging errors because no timing was configured.

This raise is just reinforcing the correct behavior and an improvement.

instro/daq/daq.py
maxleblang: We explicitly set `self._background_config.interval = 0` in `InstroDAQ.__init__()` after calling the Instrument initialization with the `background_config` arg so we're guaranteed that the user can't bypass this.

I do agree that this probably isn't the best way. I can add a specific `sw_timing_configured` bool to `InstroDAQ` if you think that's a cleaner design.

instro/daq/daq.py
hoehner: The start() method (for daqmx at least) actually starts the task itself. This is why the start() method has a background argument to allow the user to not actually start the background daemon but start the task itself.

instro/daq/daq.py
hoehner: There are 4 things that can happen when we call "start()"
- HW timed w/ background daemon (already supported)
- HW timed w/o background daemon (already supported)
- SW timed w/ background daemon (this PR should address)
- SW timed w/o background daemon (also valid just like the HW timed w/o background is valid). 

If we don't have hardware timing configured, we should just assume the user wants to do a software timed acquisition and we should probably have some default read rate that can be overwritten by configuring software timed acquisition.

I *think* this solves our breaking issue while also getting us this net-new functionality. Am i missing something?

maxleblang: Currently if there's no timing configuration and the user calls `start()` (background or not) we will error out and say you need to configure some sort of timing to call start. Additionally, If the user configures a sw timing config and then calls `start(background=False)` we'll error out as well. To me, software timed w/o background daemon is just configuring and then reading, no `start()` needed. Does that reflect what you mean?

instro/daq/daq.py
maxleblang: But if the user just plans to take software timed reads, there's no reason to call `start`

instro/daq/daq.py
hoehner: For PSUs or DMMs we don't configure timing and they are also considered "software timed". We have a default daemon rate for those instruments. We should do something similar here.

You can call start without configuring software timing and we just run the daemon at a default rate (1Hz). 

Agreed that software timing without background daemon is just calling open() and then reading/writing.

maxleblang: Ok that makes sense. So we're saying if the use does no timing configuration at all and then calls `start()` we want to start up a sw timed background daemon running at some default rate? Doing no timing configuration and calling `start(background=False)` should still raise though.

instro/daq/daq.py
hoehner: Probably true - double check that a DAQmx software timed task doesn't explicitly need to have its task commanded to "start". If that is the case then we can skip out on requiring the start() method when doing a software timed task with no background daemon. 

A software timed acquisition with a background daemon shouldn't raise.

instro/daq/daq.py
hoehner: I think we should just warn in the case where we do no timing configuration + start(background=False). It is wrong, but its harmless and shouldn't stop the script. A warning or a log message is completely sufficient here. 

The start can just be a no-op and we can say "hey friendly reminder that you're doing a software timed acquisition and have explicitly stated that you don't want a background daemon, you don't need to do that".

instro/daq/__init__.py
```diff
@@ -8,19 +8,23 @@
...
     DAQDriverBase,
     HWTimestamper,
     HWTimingException,
     InstroDAQ,
+    TimingConfigException,
 )
 from instro.daq.types import ChannelType, DAQVendor, RelayChannel
 
 __all__ = [
+    "DEFAULT_SW_SAMPLE_RATE",
```
nhschwab: Is there any reason we are including this in the public API besides wanting to import it in the unit tests?

instro/daq/daq.py
```diff
@@ -25,13 +26,16 @@
...
-from instro.lib import Instrument, InstrumentNotOpenError, Measurement
+from instro.lib import InstroError, Instrument, InstrumentNotOpenError, Measurement
 from instro.lib.instrument import publish_command, publish_measurement
 from instro.lib.publishers import Publisher
 from instro.lib.types import Command
 
 logger = logging.getLogger(__name__)
 
+# Software-timed polling rate used by start() when no AI timing was configured.
+DEFAULT_SW_SAMPLE_RATE = 1.0
```
nhschwab: why 1 second?

instro/daq/daq.py
```diff
@@ -876,16 +915,70 @@ def configure_ai_sample_rate(
...
+            )
+        if sample_rate <= 0:
+            raise ValueError(f"Software-timed sample rate must be greater than 0 Hz, got {sample_rate}.")
+
+        # Software timing needs a background loop period.
+        self._background_config.interval = 1 / sample_rate
+        self._is_sw_timing_configured = True
+
+        # Set buffer length to 10 seconds or the default Instrument length, whichever is greater
+        self._channel_buffer_length = max(int(sample_rate * 10), self._channel_buffer_length)
```
nhschwab, line 937: why 10 seconds?

instro/daq/daq.py
```diff
@@ -876,16 +915,70 @@ def configure_ai_sample_rate(
...
+        logger.info("Configured AI software timing on DAQ '%s' at %s Hz", self.name, sample_rate)
+
     def start(self, background: bool = True, **kwargs):
-        """Start hardware-timed acquisition.
+        """Start acquisition: hardware-timed, or the software-timed daemon when SW timing is configured.
+
+        With no AI timing configured, ``background=True`` falls back to software timing at
+        ``DEFAULT_SW_SAMPLE_RATE`` Hz.
```
nhschwab: Should we include the actual rate in the docstrings to make it easier for users?

instro/daq/daq.py
```diff
@@ -876,16 +915,70 @@ def configure_ai_sample_rate(
...
         self._require_open()
+        if not self.is_hw_timing_configured and not self.is_sw_timing_configured:
+            if not background:
+                # Raise when trying to do sw timed with no background daemon
+                raise TimingConfigException(
+                    f"Cannot start DAQ '{self.name}' with start(background=False) without AI timing configured. "
+                    "Call configure_ai_hw_sample_rate() first, or use start(background=True) to poll at the "
+                    f"default {DEFAULT_SW_SAMPLE_RATE} Hz software-timed rate."
+                )
+            # If no timing configrued and start called, resort to sw timed daemon at defualt rate
```
nhschwab:
```suggestion
            # If no timing configured and start called, resort to sw timed daemon at default rate
```

instro/daq/daq.py
```diff
@@ -855,12 +870,36 @@ def configure_ai_sample_rate(
...
+            sample_rate=sample_rate,
+            samples_per_channel=samples_per_channel,
+            **kwargs,
+        )
+
+    def configure_ai_hw_sample_rate(
+        self,
+        sample_rate: float,
+        samples_per_channel: int | None = None,
+        **kwargs,
```
nhschwab: `**kwargs**` is dead-code and gets dropped both here and in `configure_ai_sw_sample_rate`. This is just pattern matching against the existing `configure_ai_sample_rate` which is the source of the bug. We should definitely remove it from the two new methods. Removing it from the pre-existing method is technically a breaking change but probably a justifiable one

instro/daq/daq.py
hoehner: We picked this in a chat yesterday. Seemed reasonable as a default (like the other instruments having a sensible default).

instro/daq/daq.py
nhschwab: What do you mean by usually doesn't spam that much? It's dependent on the requested sampling rate and the OS scheduling jitter. 

Stepping back from the warning issue, I think we can actually delete the entire period check. The interval time and work time is already handled by the instrument daemon. If the issue is discoverability of overrun, then we should add a log line in the background daemon in the instrument itself (we can open a separate issue and PR for that if it doesn't already exist). Thoughts?

maxleblang: Me and John talked about this and I think that makes the most sense as well. I'll remove from here and open a new PR.

instro/daq/daq.py
maxleblang: Ok I'll just have it warn and then just return

instro/daq/daq.py
maxleblang: DAQmx doesn't need start to read the values from the configured chanels

instro/daq/__init__.py
maxleblang: Only used for unit tests

instro/daq/daq.py
maxleblang: This is the same default we use for hw timed. Just mirroring here.

instro/daq/__init__.py
maxleblang: This doesn't make sense though so I'll fix it and make it a property of `InstroDAQ` instead

cartercanedy:
@greptileai

instro/daq/daq.py
```diff
@@ -876,16 +912,74 @@ def configure_ai_sample_rate(
...
         """
         self._require_open()
+        if not self.is_hw_timing_configured and not self.is_sw_timing_configured:
+            if not background:
+                # Nothing would pace the reads, so start nothing
+                logger.error(
+                    "Calling start(background=False) without AI timing configured is unnecessary. "
+                    "Call read_analog() directly instead."
+                )
+                return
```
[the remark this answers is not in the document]
maxleblang: This was discussed and intentional. Not a bug.

maxleblang:
Closes #305

tests/daq/ni/test_nidaq_hw_and_sw_timed_hardware.py
```diff
@@ -0,0 +1,184 @@
...
+        self.assertLessEqual(
+            hw_period,
+            EXPECTED_HW_PERIOD_S + PERIOD_TOLERANCE_S,
+            msg=f"HW daemon ran at {hw_period:.4f} s/cycle, expected at most {EXPECTED_HW_PERIOD_S:.4f} s",
+        )
+
+    # =====================================================================
+    # Unsustainable software-timed rate warns
+    # =====================================================================
+    def test_sw_timed_rate_warns_when_read_outlasts_the_period(self):
```
nhschwab: Needs to be updated

docs/guides/instrumentation/daq.mdx
```diff
@@ -43,19 +43,26 @@ Channel aliases are used as channel names when data is published to Nominal Core
...
+- Use case: Low-frequency continuous monitoring without a hardware sample clock
+
 **Hardware-Timed (Buffered Acquisition)**
-- Configure sample rate with `configure_ai_sample_rate()`
+- Configure sample rate with `configure_ai_hw_sample_rate()`
 - Call `start()` to begin background acquisition
 - Data automatically published via the background daemon, OR manually fetch with `read_analog()`
 - Use case: Mid to high frequency continuous monitoring with deterministic sample timing.
 
+Hardware and software timing are mutually exclusive on one instance: configuring the second mode raises `HWTimingException`. To run both against one device, build a separate `InstroDAQ` per mode with non-overlapping channels: see the [NI hardware- and software-timed example](/instrumentation/examples/daq/daq_hw_and_sw_timed_ni).
```
nhschwab:
```suggestion
Hardware and software timing are mutually exclusive on one instance: configuring the second mode raises `TimingConfigException`. To run both against one device, build a separate `InstroDAQ` per mode with non-overlapping channels: see the [NI hardware- and software-timed example](/instrumentation/examples/daq/daq_hw_and_sw_timed_ni).
```

maxleblang:
All LabJack, NI, and MCC hardware tests pass for this change
Screen Shot 2026-08-06 at 01 50 26 161 PM
Screen Shot 2026-08-06 at 01 54 50 923 PM
Screen Shot 2026-08-06 at 01 57 20 57 PM
Screen Shot 2026-08-06 at 02 01 00 469 PM
Screen Shot 2026-08-06 at 02 10 31 496 PM

Note: LabJack T8 failure is due to a separate identified bug (#360) and fails on `main` as well so this is not blocking for this PR

hoehner:
Testing on Keysight 34980A using the hardware -m test script
```
============================= test session starts =============================
platform win32 -- Python 3.13.14, pytest-9.1.1, pluggy-1.6.0 -- C:\development\instro\.venv\Scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\development\instro
configfile: pyproject.toml
collecting ... collected 10 items

..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_01_device_info_and_idn          *IDN? = Agilent Technologies,34980A,MY44001757,2.41-2.11-2.07-1.05
         SCPI version = 1994.0
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_02_sw_timed_analog_read          4002 (sw-timed) = 1.994500 V
         4002 (sw-timed) = 1.993900 V
         4002 (sw-timed) = 1.994500 V
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_03_relay_close_open_roundtrip          after close: ROUT:CLOS? (@4003) = 1
         after open:  ROUT:OPEN? (@4003) = 1
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_04_multi_channel_scan          scanned channels 4002, 4021 with no SCPI error
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_05_hw_timed_scan          fetched 5 samples at 10.0 Hz
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_06_buffer_depth_telemetry          points_in_buffer telemetry = 1.0
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_07_digital_line_loopback          8101/0<-0 | 8201/0=0
         8101/0<-1 | 8201/0=0
         8101/0<-0 | 8201/0=0
         8101/0<-1 | 8201/0=0
         8101/0<-0 | 8201/0=0
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_08_error_queue_clean          SCPI error queue clean after analog + relay ops
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_09_analog_output_unsupported          analog output correctly rejected (no AO path on 34980A)
PASSED
..\..\tests\daq\keysight34980A\test_keysight_34980A_hardware.py::TestKeysight34980AHardware::test_10_digital_port_write_read          8101<-0x00 | 8201=0xB8
         8101<-0x01 | 8201=0xB8
         8101<-0xFF | 8201=0xB8
         8101<-0xAA | 8201=0xB8
         8101<-0x00 | 8201=0xB8
PASSED

============================= 10 passed in 8.04s ==============================
```

hoehner:
@nhschwab and @maxleblang - I ran the example script on the Keysight DAQ we have here in LA as well as the hardware tests for this driver. Both demonstrated that this driver is behaving as expected using the software timed daemon functionality.

nhschwab (approved):

PR #335, opened 2026-07-31
feat(awg): add modulation to AWG

LizzyC-115:
## Summary

Adding modulation feature to AWG drivers.

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification
Connection to Tektronix Scope:
Screenshot 2026-07-31 at 4 50 09 PM

Hardware Tests:
Screenshot 2026-07-31 at 4 50 26 PM

Software Tests:
Screenshot 2026-07-31 at 4 50 49 PM

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for Reviewers

For this implementation, we are decoupling setting modulation features (mod_type, shape, and magnitude) from enabling modulation on a specific channel. This is to avoid having the user reset the channel's modulation in procedural tasks. Therefore, instead of solely modulate(), we use enable_modulation(), set_modulation(), and disable_modulation(). 

This pattern-matches well with Keysight 33500B Waveform Generator as well. For the Keysight, you can enable/disable modulation with the command `[mod_type]:STAT:ON` or `[mod_type]:STAT:OFF`.


packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +188,35 @@ def get_output_load(self, channel: int) -> float | None:
     def align_phase(self) -> None:
         self._visa.write(":SOUR1:PHAS:SYNC")
 
+    def modulate(self, channel: int, mod_type: ModulationType, shape: Waveform, magnitude: float) -> None:
+        _check_channel(channel)
+        if not isinstance(mod_type, ModulationType):
+            raise TypeError(f"mod_type must be a ModulationType, got {type(mod_type).__name__}")
+        carrier = _mod_carrier(shape)
```
nhschwab: This naming is backwards and confusing. In modulation, there are two signals. The carrier is the usually higher frequency, stable waveform. In the case of InstroAWG, the carrier is being set with `set_waveform`. The modulating signal or modulator is the usually lower frequency, baseband signal that encodes information onto the carrier by changing the amplitude, frequency, or phase of the carrier. So in this case, the variable `carrier` is actually the modulator, and the method `_mod_carrier` is actually validating the modulating waveform. Let's rename these and document it to make it clear which is which.

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +188,35 @@ def get_output_load(self, channel: int) -> float | None:
     def align_phase(self) -> None:
         self._visa.write(":SOUR1:PHAS:SYNC")
 
+    def modulate(self, channel: int, mod_type: ModulationType, shape: Waveform, magnitude: float) -> None:
```
nhschwab: The magnitude argument is a little misleading. For AM, the magnitude becomes amplitude depth as a percent, for FM it becomes frequency deviation in Hz, for PM it becomes phase deviation in degrees, for ASK it becomes second amplitude in Vpp, and for FSK it becomes the hop frequency in Hz. We should document this somewhere for developers and users

nhschwab: Also worth noting for ASK and FSK the shape argument is silently ignored. Might be worth adding docs just to flag to the user that the argument doesn't do anything for those mod types

nhschwab:
My only major concern is that the modulation command as currently implemented is not stateful. In other words, a user can send a command to modulate a waveform, but they can't turn it off or turn if back on again after the fact. I think the correct shape here is to have an internally tracked struct for each channel that records the modulation type and its characteristics and to have methods for updating the modulation signal and for turning on/off modulation for a given channel

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
LizzyC-115: Updated to have two helper functions: _validate_modulator and _validate_carrier to catch invalid calls. Also added a comment to differentiate between carrier and modulator in the function.

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
LizzyC-115: Added more information in the docstring for awg.py to explain what each type uses `magnitude` for. Also added a section in awg.mdx to explain what is valid, with a table.

LizzyC-115:
Added `ModulationState` as a class to the Rigol driver to allow enable and disable functions that the user can call. Each channel has a dedicated instance. The class keeps track of type, shape, magnitude, and whether or not the channel has been enabled. We clear these states to None when modulation is disabled to avoid leaks.

LizzyC-115:
Decoupled modulate() into enable_modulation() and set_modulation()

LizzyC-115:
@claude review

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -87,6 +88,18 @@ def align_phase(self) -> None:
...
+        """Enable modulation on the given channel using its last-configured set_modulation() settings."""
+        raise NotImplementedError(f"enable_modulation is not implemented for {type(self).__name__}")
+
+    def set_modulation(self, channel: int, mod_type: ModulationType, shape: Waveform, magnitude: float) -> None:
+        """Configure channel's carrier modulation with modulator shape; call enable_modulation() to activate it."""
+        raise NotImplementedError(f"set_modulation is not implemented for {type(self).__name__}")
+
+    def disable_modulation(self, channel: int) -> None:
+        """Disable modulation on the given channel, if enabled; the configured settings are kept for re-enabling."""
+        raise NotImplementedError(f"disable_modulation is not implemented for {type(self).__name__}")
```
nhschwab: Let's collapse `enable_modulation` and `disable_modulation` into one method that takes the channel and a boolean `enable` argument, just like the `output_enable` method. This follows the precedent set for the waveform setting and simplifies command publishing. Let's name the method `modulation_enable` for consistency

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -351,3 +364,44 @@ def get_output_load(self, channel: int, **kwargs) -> Measurement | None:
...
+    @publish_command
+    def disable_modulation(self, channel: int, **kwargs) -> Command:
+        """Disable modulation on the given channel, if enabled; the configured settings are kept for re-enabling."""
+        self._check_channel(channel)
+        with self._resource_lock:
+            self._driver.disable_modulation(channel=channel)
+            timestamp = time.time_ns()
+            self._check_errors()
+        descriptor = f"ch{channel}.modulation.cmd"
+        return self._package_command(descriptor, "OFF", timestamp, **kwargs)
```
nhschwab: Same comment as above, collapse into one `modulation_enable` method with channel and boolean as args

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated! Made one `modulation_enable()` and updated all instances of modulation in tests and docs

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +199,61 @@ def get_output_load(self, channel: int) -> float | None:
...
+    def set_modulation(self, channel: int, mod_type: ModulationType, shape: Waveform, magnitude: float) -> None:
+        _check_channel(channel)
+        if not isinstance(mod_type, ModulationType):
+            raise TypeError(f"mod_type must be a ModulationType, got {type(mod_type).__name__}")
+        # `shape` is the modulator/baseband signal, the carrier is what set_waveform last programmed.
+        state = self._modulation_states[channel]
+        if state.enabled:
+            raise RuntimeError(
+                f"channel {channel} already has modulation enabled; disable it via modulation_enable() first"
+            )
```
nhschwab: why do we want to block on changing the modulator of a channel if it is enabled? Does the Rigol error if you try to do this? I don't see any reason to disallow the user from doing this

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -351,3 +360,33 @@ def get_output_load(self, channel: int, **kwargs) -> Measurement | None:
...
+    @publish_command
+    def modulation_enable(self, channel: int, enable: bool, **kwargs) -> Command:
+        """Enable or disable modulation on the given channel."""
+        self._check_channel(channel)
+        with self._resource_lock:
+            self._driver.modulation_enable(channel=channel, enable=enable)
+            timestamp = time.time_ns()
+            self._check_errors()
+        descriptor = f"ch{channel}.modulation.cmd"
+        return self._package_command(descriptor, "ON" if enable else "OFF", timestamp, **kwargs)
```
nhschwab: I think we want the command values to be bools (i.e. `enable`) instead of these arbitrary strings. See how PSU does it for publishing enabled commands

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
LizzyC-115: From the way the code is structured, it was to prevent leaving the old modulation on. If we did `AM:STAT:ON` then changed to FM, we'd leave AM on. But I don't think this errors on the Rigol so we can change it and add a docstring to make it clear to the user. 

UPDATE(BLUF):  issue is resolved by removing the enabled check in `modulation_enable()`. Removing this line and leaving out the docstring

packages/instro-unstable/instro/unstable/awg/awg.py
nhschwab: Yeah see how you already do it for `InstroAWG.output_enable()`

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +199,61 @@ def get_output_load(self, channel: int) -> float | None:
...
+            state.modulation_type = mod_type
+            state.modulation_shape = shape
+            state.modulation_magnitude = magnitude
+
+    def modulation_enable(self, channel: int, enable: bool) -> None:
+        _check_channel(channel)
+        with self._visa.lock():
+            state = self._modulation_states[channel]
+            if state.enabled == enable:
+                return
```
nhschwab: I don't think we want this guard, the commands are idempotent on the instrument and help avoid state mismatches between software and hardware. Let's remove this so we send the actual command regardless of software state

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +199,61 @@ def get_output_load(self, channel: int) -> float | None:
...
+            elif mod_type is ModulationType.PSK:
+                self._visa.write(f":SOUR{channel}:PSK:SOUR INT")
+                self._visa.write(f":SOUR{channel}:PSK:INT:RATE {frequency_hz}")
+                self._visa.write(f":SOUR{channel}:PSK:PHAS {magnitude}")
+            else:
+                raise AssertionError(f"unhandled ModulationType {mod_type}")
+
+            state.modulation_type = mod_type
+            state.modulation_shape = shape
+            state.modulation_magnitude = magnitude
```
nhschwab: Do we check visa errors at all in this method before saving state? My only concern is a command that gets rejected by the instrument but we persist the config in software state

LizzyC-115: Added check_errors()

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
LizzyC-115: Removed

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated to package the enable parameter instead

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +188,54 @@ def get_output_load(self, channel: int) -> float | None:
...
+            self.check_errors()
+
+    def get_modulation_state(self, channel: int) -> tuple[ModulationType, bool]:
+        """Get the modulation type and enabled state currently active on channel, read from the instrument."""
+        _check_channel(channel)
+        with self._visa.lock():
+            mod_type = ModulationType(self._visa.query(f":SOUR{channel}:MOD:TYP?").strip())
+            enabled = self._visa.query(f":SOUR{channel}:MOD:STAT?").strip() == "ON"
+        return mod_type, enabled
+
```
LizzyC-115: Choosing to defer to hardware to query state instead of having a ModulationState class. This was to avoid inconsistencies with software state and hardware state. In the previous iteration, we only used type and enabled fields, so that is what we query for here.

packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
```diff
@@ -180,6 +188,54 @@ def get_output_load(self, channel: int) -> float | None:
...
+                self._visa.write(f":SOUR{channel}:FSK:SOUR INT")
+                self._visa.write(f":SOUR{channel}:FSK:INT:RATE {frequency_hz}")
+                self._visa.write(f":SOUR{channel}:FSK {magnitude}")
+            elif mod_type is ModulationType.PSK:
+                self._visa.write(f":SOUR{channel}:PSK:SOUR INT")
+                self._visa.write(f":SOUR{channel}:PSK:INT:RATE {frequency_hz}")
+                self._visa.write(f":SOUR{channel}:PSK:PHAS {magnitude}")
+            else:
+                raise AssertionError(f"unhandled ModulationType {mod_type}")
+            self._visa.write(f":SOUR{channel}:MOD:TYP {mod_type.value}")
```
LizzyC-115, line 219: This automatically disables the previous modulation type and enables the new modulation type

LizzyC-115: Avoids having a check to disable before setting modulation and allows `modulation_enable` to be idempotent

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -87,6 +88,18 @@ def align_phase(self) -> None:
...
+        """Configure channel's carrier modulation with modulator shape; call modulation_enable() to activate it."""
+        raise NotImplementedError(f"set_modulation is not implemented for {type(self).__name__}")
+
+    def modulation_enable(self, channel: int, enable: bool) -> None:
+        """Enable or disable modulation on the given channel."""
+        raise NotImplementedError(f"modulation_enable is not implemented for {type(self).__name__}")
+
+    def get_modulation_state(self, channel: int) -> tuple[ModulationType, bool]:
+        """Get the modulation type and enabled state currently active on channel, read directly from the instrument."""
+        raise NotImplementedError(f"get_modulation_state is not implemented for {type(self).__name__}")
```
nhschwab: Any reason this shouldn't be split into two separate methods, `get_modulation_type` and `get_modulation_state`? Follows the pattern already set by `get_waveform` and `get_output_state`, is a cleaner, simpler API, and makes the measurement publishing more straight forward

LizzyC-115: Split these functions into two. `get_modulation_type()` should return the waveform type that we've previously set on the channel and `get_modulation_state()` should return a boolean on whether we've enabled modulation for that channel

tests/unstable/awg/rigol/test_rigol_dg1022z_hardware.py
```diff
@@ -1,4 +1,9 @@
-"""Rigol DG1022Z hardware smoke tests."""
+"""Rigol DG1022Z hardware smoke tests.
+
+One to two tests per feature: one valid sequence, one invalid/error-check sequence. Where no
```
LizzyC-115, line 3: Note: the changes in this file are just refactoring the tests to be more compact (i.e. grouping like tests together). It shouldn't affect functionality compared to before this PR

tests/unstable/awg/rigol/test_rigol_dg1022z_software.py
```diff
@@ -10,6 +10,7 @@
 from instro.unstable.awg.types import (
     AmplitudeMeasurementUnit,
     Arbitrary,
+    ModulationType,
     Pulse,
     Sawtooth,
```
LizzyC-115, line 15: Note: the changes in this file are just refactoring the tests to be more compact (i.e. grouping like tests together). It shouldn't affect functionality compared to before this PR

tests/unstable/awg/test_awg.py
```diff
@@ -1,24 +1,21 @@
-"""Tests for AWGDriverBase contract, waveform definitions, and InstroAWG composition."""
+"""Tests for AWGDriverBase contract and InstroAWG's generic wiring to a driver.
+
+Scoped to what's actually generic about the abstraction layer: the driver contract, lifecycle,
+the shared check_errors/tagging conventions, and channel-bounds enforcement across every method.
+Per-method roundtrip/packaging behavior (set_waveform, amplitude, offset, output, modulation, ...)
```
LizzyC-115, line 5: Note: the changes in this file are just refactoring the tests to be more compact (i.e. grouping like tests together). It shouldn't affect functionality compared to before this PR

nhschwab (approved):

PR #336, opened 2026-08-01
feat(lib): transport base class that implements shared ownership of transport

wharwood:
## Summary

Adds `TransportBase` in `instro/lib/transports/transport_base.py`, a base class that lets several owners share one `VisaDriver` or `ModbusDriver` connection. `open(holder)` opens the connection if needed and returns `True` only for the first owner, so one-time device setup (taking a combined instrument's remote lock, say) runs exactly once. `close(holder)` removes that owner and tears the connection down only when the last one leaves. Both transports now inherit the holder list, the reentrant lock, and the `open`/`close`/`__del__` lifecycle instead of duplicating them, keeping the guarded bare `close()` that declines with a warning while any owner remains and the `__del__` GC backstop that bypasses the guard rather than strand a resource. Ownership is **two-level**. A device that serves several categories holds the transport once on its own behalf and tracks its own category views, so device-level teardown runs in the device's release path while the session is still up, before it calls `transport.close(self)`. That keeps each "last release" unambiguous: the device's last view leaving triggers device teardown, the transport's last holder leaving closes the connection. This unblocks combined instruments like the EA PSB 10000 series (#379), where a PSU-shaped driver and an ELoad-shaped driver share one physical connection without either closing it out from under the other. Closes INSTRO-512.

## Type of change

- [ ] Bug fix (`fix`)
- [X] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

This was tested with the added unit tests and mock transports.

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

This is the basis for drivers that will share a transport but implement more than one base instrument. It should be thoroughly discussed prior to merge.


wharwood:
@greptile @claude can we rereview this code?

hoehner (dismissed):

instro/lib/transports/transport_base.py
```diff
@@ -0,0 +1,95 @@
...
+            if self._holder_index(holder) is not None:
+                return False
+            self._open_session()
+            self._holders.append(holder)
+            return len(self._holders) == 1
+
+    def close(
+        self,
+        holder: object | None = None,
+        on_last_release: Callable[[], None] | None = None,
```
hoehner: Is it worth putting this on the "open()" method instead? This argument on the close() method makes session teardown dependent on close order (only the last closers callback will actually run). If the last closer just so happens to forget to add the callback, then we never truly close the session. If we put it on the open() the transport base will store the callback and we'll get the "automatic close" functionality on the guaranteed last closer for free.

instro/lib/transports/transport_base.py
```diff
@@ -0,0 +1,95 @@
...
+    ) -> None:
+        """Close and teardown; with a holder, remove it and tear down only when the last owner leaves."""
+        with self._lock:
+            if holder is not None:
+                index = self._holder_index(holder)
+                if index is None:
+                    return
```
hoehner, line 61: Lets raise here - this is an unexpected situation where the user expects there to have been a session in "index" but in reality there is nothing there. We should throw something saying "yo, this isn't a registered holder of this transport" or something similar.

hoehner (changes requested):
Two minor issues worth digging into

hoehner:
Looks like we have one quick conflict to resolve as well.

instro/lib/transports/transport_base.py
wharwood: You thinking a list of callbacks? Last close iterates through everything?

wharwood:
@claude @greptile rereview.

hoehner (approved):

PR #338, opened 2026-08-03
chore(instrument): log warning when background daemon cant run at requested rate

### LINKED ISSUE #337: chore(instrument): log warnings when background daemon can't run at requested rate
We currently let the user configure the background daemon interval at any rate they want. However, the real rate the background daemon runs at is bounded by the work time of the daemon. Therefore, we should warn the user if their requested rate can't be accomplished.

We should average the work time of 10 cycles and compare it to the requested interval. If the average worktime is higher than the requested interval, warn the user once that their requested interval can't be achieved and log the max interval achievable based on the average work time.

maxleblang:
## Summary

Based on the average work time of 10 cycles of the background daemon, warn the user if their requested rate isn't achievable (avg_work_time > requested_rate). We only warn once per instrument run.

## Type of change

- [ ] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

ver

Entire test suite passes

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

Because of the way we pace the loop by calling `wait()` and the GIL, the daemon loop often runs at a slightly slower rate than the requested interval (loop_time = work_time + sw_latency). This is a pitfall of a sw timed loop in Python that we can't really get around. I just wanted to bring attention to the fact that this is only checking if the work time of the daemon is higher than the requested rate, not that the daemon is delivering on the requested loop interval (loop_time > requested_interval?)

### DIFF
--- a/instro/lib/instrument.py
+++ b/instro/lib/instrument.py
@@ -100,7 +100,7 @@ class Instrument:
         self._background_stop_event = threading.Event()
         self._background_methods: list[tuple[Callable, tuple, dict]] = []
         # Used to determine whether requested interval is achievable or not
-        self._daemon_loop_times: deque[float] = deque(maxlen=10)
+        self._daemon_work_times: deque[float] = deque(maxlen=10)
         self._interval_warning_issued = False
 
         self._channel_buffer_length: int = 10
@@ -242,7 +242,7 @@ class Instrument:
 
     def _reset_daemon_timing(self):
         """Drop the timing window and re-arm the warning so cycles are judged against the current interval."""
-        self._daemon_loop_times.clear()
+        self._daemon_work_times.clear()
         self._interval_warning_issued = False
 
     def add_background_daemon_function(self, method: Callable, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


instro/lib/instrument.py
```diff
@@ -234,6 +238,12 @@ def background_interval(self):
     @background_interval.setter
     def background_interval(self, seconds: float):
         self._background_config.interval = seconds
+        self._reset_daemon_timing()
+
+    def _reset_daemon_timing(self):
```
nhschwab, line 243: We should also call this method whenever the background daemon is redefined by the user

instro/lib/instrument.py
```diff
@@ -360,12 +372,37 @@ def _background_daemon_loop(self):
...
+
+        This doesn't give timing guarantees about the background daemon rate,
+        just warn when work_time > requested interval.
+        """
+        self._daemon_work_times.append(work_time_s)
+        if self._interval_warning_issued or len(self._daemon_work_times) != self._daemon_work_times.maxlen:
+            return
+
+        average_work_time_s = sum(self._daemon_work_times) / len(self._daemon_work_times)
+        if average_work_time_s <= self._background_config.interval:
```
nhschwab, line 401: This check will always fail for InstroDAQ when it has the default 0 sec background interval for hardware-paced reads. We should short-circuit for interval <=0

instro/lib/instrument.py
```diff
@@ -360,12 +372,37 @@ def _background_daemon_loop(self):
...
+            return
+
+        logger.warning(
+            "Background daemon for instrument '%s' cannot achieve the requested interval of %.6f s; "
+            "average work time over %d cycles is %.6f s, which is the fastest achievable interval.",
```
nhschwab: Technically not true that it is the fastest achievable interval, can we change the wording to be more accurate? Instead of saying it is the fastest achievable interval, we can provide guidance on how to fix? "Set background daemon interval to a value greater than the work time" or something like that

instro/lib/instrument.py
maxleblang: Good catch

instro/lib/instrument.py
```diff
@@ -360,12 +373,40 @@ def _background_daemon_loop(self):
...
+        """
+        # An interval of 0 asks for free-running, so there is no rate to fall short of.
+        if self._background_config.interval <= 0:
+            return
+
+        self._daemon_work_times.append(work_time_s)
+        if self._interval_warning_issued or len(self._daemon_work_times) != self._daemon_work_times.maxlen:
+            return
+
+        average_work_time_s = sum(self._daemon_work_times) / len(self._daemon_work_times)
```
nhschwab: Is it possible that `len(self._daemon_work_times) == 0` due to a reset between the previous check? We might want to grab a static copy of `self._daemon_work_times` in a local variable just in case

maxleblang: I don't understand the scenario you're talking about. We should never be able to reach the division if `self._daemon_work_times` isn't 10 (`maxlen` in this case). What would grabbing a static topic buy us?

nhschwab: It's technically possible that `self._reset_daemon_timing()` gets called in between the check on line 394 and computation of `average_work_time_s` on line 397., in which case `self._daemon_work_times` becomes empty with length 0. I'm not seeing any guarantees of protection against that.

maxleblang: I understand now. I'll have reset just create a new `deque` pointer instead of clearing the shared one.

nhschwab (approved):

PR #341, opened 2026-08-03
feat(lib): add VISA instrument discovery API, consolidate CLI onto it

dalvarez-204:
Screenshot 2026-08-03 at 11 08 32 AM
Screenshot 2026-08-03 at 11 08 53 AM
Screenshot 2026-08-03 at 11 10 43 AM

Adds `scan_visa_resources()` and its result dataclasses (`VisaScanResult`, `VisaInstrumentInfo`, `VisaScanError`, `VisaUnrecognizedInstrument`) to `instro.lib` as a public, programmatic VISA discovery API. `instro discover` (the CLI) is refactored onto this shared implementation instead of duplicating its own scan loop and `_IDN_MAP`.

(added a new ERRORS table to the CLI, since errors are now structured data returned after the scan rather than printed live during it).

Degraded-interface diagnostics and serial-port listing stay CLI-only.
- `scan_visa_resources()` accepts an already-open `ResourceManager` (`rm=`).
- `VisaScanError` gained an optional `hint` field: `message` stays the raw `str(exc)` unconditionally, so a consumer that just wants results isn't forced through an interpretation layer; `hint` carries the CLI's friendlier guidance ("permission denied - check udev rules", "USB backend missing - install libusb") only when a known exception pattern matches, for callers that want to display it.

Docs: added a `scan_visa_resources` reference page (`docs/reference/src/reference/discover.md`) and a "Discovering programmatically" section to the CLI guide.

## Known limitation

VISA's resource enumeration (`pyvisa`'s `list_resources()`) can't auto-discover a raw TCP-socket instrument (e.g. the bundled PSU simulator) -- there's no broadcast/mDNS mechanism for a plain socket resource. Discovery only works for instruments VISA can natively enumerate (USB, GPIB, some LAN/VXI-11 instruments). This is a pre-existing limitation flagged by a `TODO` in the module, not something introduced here.

## Test plan
- [x] `just check` and `just test` pass
- [x] Ran `python -m instro.cli.main discover` for real (no devices connected) -- clean output, no crash


instro/lib/discover.py
```diff
@@ -0,0 +1,154 @@
...
+    ("KEYSIGHT TECHNOLOGIES", "34461A"): ("dmm", "Keysight34461A", None, None),
+    ("AGILENT TECHNOLOGIES", "34461A"): ("dmm", "Keysight34461A", None, None),
+    ("B&K PRECISION", "9115"): ("psu", "BK9115", "bk_9115", 1),
+    ("B&K PRECISION", "9140"): ("psu", "BK914X", "bk_914x", 3),
+    ("RIGOL TECHNOLOGIES", "DP811"): ("psu", "RigolDP800", "rigol_dp800", 1),
+    ("RIGOL TECHNOLOGIES", "DP821"): ("psu", "RigolDP800", "rigol_dp800", 2),
+    ("RIGOL TECHNOLOGIES", "DP831"): ("psu", "RigolDP800", "rigol_dp800", 3),
+    ("RIGOL TECHNOLOGIES", "DP832"): ("psu", "RigolDP800", "rigol_dp800", 3),
+    ("SIGLENT TECHNOLOGIES", "SPD3303"): ("psu", "SiglentSPD3303", "siglent_spd3303", 3),
+    ("GENESYS", "GEN"): ("psu", "TDKLambdaGenesys", "tdk_lambda_genesys", 1),
```
dalvarez-204: remove this!

instro/lib/discover.py
```diff
@@ -0,0 +1,153 @@
...
+    ("B&K PRECISION", "BK85"): ("eload", "BK85XXB", None, None),
+    ("KEYSIGHT TECHNOLOGIES", "DSOX120"): ("scope", "Keysight1200X", "keysight_1200x", 2),
+    ("KEYSIGHT TECHNOLOGIES", "EDUX105"): ("scope", "Keysight1200X", "keysight_1200x", 2),
+    ("TEKTRONIX", "MSO22"): ("scope", "Tektronix2SeriesMSO", "tektronix_2series", 4),
+    ("TEKTRONIX", "MSO24"): ("scope", "Tektronix2SeriesMSO", "tektronix_2series", 4),
+    ("SIGLENT TECHNOLOGIES", "SDS1104X-E"): ("scope", "SiglentSDS1000XE", "siglent_sds1000x_e", 4),
+    ("SIGLENT TECHNOLOGIES", "SDS1202X-E"): ("scope", "SiglentSDS1000XE", "siglent_sds1000x_e", 2),
+    ("SIGLENT TECHNOLOGIES", "SDS1204X-E"): ("scope", "SiglentSDS1000XE", "siglent_sds1000x_e", 4),
+}
+
```
dalvarez-204, line 69: new _IDN_MAP values have been tested in another PR

instro/lib/discover.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    if isinstance(exc, pyvisa.errors.VisaIOError) and "SYSTEM_ERROR" in str(exc):
+        return "permission denied - check udev rules"
+    err_str = str(exc)
+    if "No backend available" in err_str or "PyUSB" in err_str:
+        return "USB backend missing - install libusb"
+    return None
+
+
+def scan_visa_resources(
```
dalvarez-204, line 86: scan functionality is the same as in original CLI work

instro/psu/psu.py
```diff
@@ -243,6 +269,34 @@ def _execute_measurement(
...
+    def discover(
+        cls,
+        backend: str | None = None,
+        timeout: int = 2,
+    ) -> list[PSUConfig]:
+        """Scan VISA resources and return PSUConfig for each recognized PSU."""
+        from instro.lib.discover import scan_visa_resources
+        from instro.lib.transports.visa import VisaConfig
+        from instro.lib.types import DeviceInfo
+        from instro.psu.config import VisaDriverConfig
```
dalvarez-204: these might be able to be hoisted out

instro/lib/discover.py
nhschwab, line 1: Seems like `scan_visa_resources` and the dataclasses should be exported as part of the instro.lib API

nhschwab:
Add docs for lib level discovery method and relevant types

instro/psu/psu.py
```diff
@@ -243,6 +246,29 @@ def _execute_measurement(
...
+                    PSUConfig(
+                        device=DeviceInfo(name=info.resource),
+                        driver=VisaDriverConfig(
+                            name=info.driver_class_name,
+                            num_channels=info.num_channels,
+                            visa=VisaConfig(visa_resource=info.resource, visa_backend=backend),
+                        ),
+                    )
+                )
+        return configs
```
nhschwab: If this is going to become a class method, we should also add it for all other instrument classes

instro/psu/psu.py
```diff
@@ -243,6 +246,29 @@ def _execute_measurement(
...
+        timeout: int = 2,
+    ) -> list[PSUConfig]:
+        """Scan VISA resources and return PSUConfig for each recognized PSU."""
+        result = scan_visa_resources(backend=backend, timeout=timeout)
+        configs = []
+        for info in result.instruments:
+            if info.category == "psu" and info.vendor_key is not None and info.num_channels is not None:
+                configs.append(
+                    PSUConfig(
+                        device=DeviceInfo(name=info.resource),
```
nhschwab: can we choose a friendlier device name? Don't want every single published channel with the visa resource string as the prefix

instro/psu/psu.py
```diff
@@ -243,6 +246,29 @@ def _execute_measurement(
             descriptor = f"ch{channel}.{channel_suffix}"
         return self._package_measurement(descriptor, val, timestamp, **kwargs)
 
+    @classmethod
+    def discover(
```
nhschwab: The more I think about it the more I'm against making this a class method. We would have to essentially copy and paste this for every instrument as we add support for configs. I think what makes more sense would be a helper function in discover that returns an array of instrument config objects for discovered instruments, but we should only plan to add that once we have config support for all instruments

instro/lib/discover.py
```diff
@@ -0,0 +1,153 @@
...
+    instruments: list[VisaInstrumentInfo]
+    unrecognized: list[VisaUnrecognizedInstrument]
+    errors: list[VisaScanError]
+
+
+# Key: (vendor, model) substrings matched case-insensitively against the *IDN? response.
+# Value: (category, driver_class_name, vendor_key, num_channels). vendor_key/num_channels are
+# only meaningful for categories with a class-name-keyed driver registry (currently psu, scope);
+# other categories (dmm, eload) don't need them to resolve a driver, so they're left None.
+_IDN_MAP: dict[tuple[str, str], tuple[str, str, str | None, int | None]] = {
```
nhschwab: Seems like `vendor_key` is dead code and unused. Unless there is a future plan for consuming it, feels like we can remove it from the map and `VisaInstrumentInfo` and all other references

nhschwab (approved):

PR #353, opened 2026-08-06
fix(lib): import exceptions from nominal.core

### LINKED ISSUE #352: CLI broken on fresh install: `from nominal import exceptions` fails with nominal>=1.143
## Symptom

A fresh \`uv tool install instro\` (any platform) crashes on every CLI invocation:

\`\`\`
File "...\instro\lib\nominal.py", line 11, in <module>
    from nominal import exceptions
ImportError: cannot import name 'exceptions' from 'nominal' (unknown location)
\`\`\`

## Root cause

\`instro/lib/nominal.py\` imports \`from nominal import exceptions\`. That path was deprecated in the nominal SDK (in 1.99.0 it is a shim that redirects to \`nominal.core.exceptions\` with a \`UserWarning\`) and was removed in **nominal 1.143.0**, where the top-level \`nominal/__init__.py\` was dropped and \`nominal\` became a namespace package (hence the \`(unknown location)\` in the error).

instro's published constraint is \`nominal>=1.99.0,<2.0.0\`, so fresh installs resolve the latest nominal (1.156.1) and fail at import. Development never sees it because \`uv.lock\` pins nominal at 1.99.0.

## Fix

Import from the canonical location, valid across the whole pinned range (verified in the 1.99.0 and 1.156.1 wheels):

\`\`\`python
from nominal.core import exceptions
\`\`\`

Also bump the locked nominal so CI exercises the post-1.143 namespace layout.

## Workaround

\`uv tool install instro --with "nominal<1.143" --force\`

nhschwab:
Fixes #352

## Problem

Every fresh install of instro (`uv tool install instro`) crashes on any CLI invocation:

```
File "...\instro\lib\nominal.py", line 11, in <module>
    from nominal import exceptions
ImportError: cannot import name 'exceptions' from 'nominal' (unknown location)
```

The top-level `nominal.exceptions` shim (deprecated since before nominal 1.99.0 in favor of `nominal.core.exceptions`) was removed in **nominal 1.143.0**, where `nominal` became a namespace package with no top-level `__init__.py`. instro's constraint `nominal>=1.99.0,<2.0.0` lets fresh installs resolve the latest nominal, which then fails at import. Development never saw it because `uv.lock` pinned nominal at 1.99.0.

## Fix

- Import from the canonical location: `from nominal.core import ... exceptions`. Verified against both the 1.99.0 and 1.156.1 wheels, so it works across the entire pinned range — no constraint change needed.
- Bump the locked nominal to 1.156.1 (pulls `nominal-streaming` 0.5.8 → 0.8.2 as its transitive requirement) so CI runs against the post-1.143 namespace layout.

## Regression coverage

No new test: the lock bump is the coverage. With nominal ≥1.143 in the lock, the old import fails at test collection — the full suite now exercises the layout that broke fresh installs. Verified locally: `just check-python` clean, `just test-python` passes (1556 passed; one unrelated failure from a local untracked example file not in the repo), and `instro --help` imports cleanly against nominal 1.156.1.

### DIFF
--- a/instro/lib/nominal.py
+++ b/instro/lib/nominal.py
@@ -8,8 +8,7 @@ from typing import Any, TypedDict
 
 import requests.exceptions
 import urllib3.exceptions
-from nominal import exceptions
-from nominal.core import Dataset, NominalClient
+from nominal.core import Dataset, NominalClient, exceptions
 from nominal.experimental.logging import install_nominal_log_handler as _install_nominal_log_handler
 
 from instro.lib.types import Measurement
--- a/uv.lock
+++ b/uv.lock
@@ -63,15 +63,6 @@ wheels = [
 ]
 
 [[package]]
-name = "cachetools"
-version = "6.2.0"
-source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/9d/61/e4fad8155db4a04bfb4734c7c8ff0882f078f24294d42798b3568eb63bff/cachetools-6.2.0.tar.gz", hash = "sha256:38b328c0889450f05f5e120f56ab68c8abaf424e1275522b138ffc93253f7e32", size = 30988, upload-time = "2025-08-25T18:57:30.924Z" }
-wheels = [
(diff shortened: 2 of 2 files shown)


dalvarez-204 (approved):
looks good. 🥇

PR #358, opened 2026-08-06
chore(ci): add scheduled latest-deps test workflow

### LINKED ISSUE #357: CI: test against latest-resolvable dependencies to catch upstream breaking changes
## Motivation

nominal-io/instro#352: nominal 1.143.0 (published 2026-05-11) removed the deprecated top-level `nominal.exceptions` shim, which made every fresh instro install crash on import for ~3 months before an end user reported it. CI never caught it because tests only run against the pinned `uv.lock` (nominal 1.99.0), while end users installing from PyPI resolve dependencies freely within the ranges in `pyproject.toml`.

Any allowed-range breaking change in an upstream dependency is currently invisible to CI until an end user hits it.

## Proposal

Add a scheduled workflow that re-resolves all dependencies to the latest versions allowed by `pyproject.toml` (`uv lock --upgrade`), installs with all extras, and runs the Python test suite plus a CLI import smoke test. This mirrors what a fresh `uv tool install instro` sees.

* Scheduled (daily) + `workflow_dispatch`, not PR-triggered — an upstream breakage shouldn't turn unrelated PRs red, and a scheduled failure notifies maintainers directly.
* Cheap: single OS, bounds-of-support Python versions only.

This would have flagged the nominal-io/instro#352 breakage within a day of nominal 1.143.0 shipping instead of three months later.

nhschwab:
Fixes nominal-io/instro#357

## Motivation

PR CI only ever tests against the pinned `uv.lock`, so a breaking release in an upstream dependency that still satisfies the `pyproject.toml` ranges is invisible until an end user installs fresh from PyPI. That's exactly how nominal 1.143.0 (removed the deprecated `nominal.exceptions` shim, 2026-05-11) broke every fresh instro install for \~3 months before being reported (nominal-io/instro#352, fixed in nominal-io/instro#353).

## What this adds

`.github/workflows/latest-deps-test.yml` — a daily scheduled (+ `workflow_dispatch`) workflow that mirrors what a fresh `uv tool install instro` resolves:

1. `uv lock --upgrade` — re-resolve everything to the latest versions `pyproject.toml` allows (lockfile is not committed back)
2. `uv sync --all-extras`
3. `uv run instro --version` — CLI entry-point smoke test (the nominal-io/instro#352 failure mode)
4. `just test-python`

Deliberately scheduled rather than PR-triggered: an upstream breakage isn't the fault of any open PR and shouldn't turn unrelated PRs red. Kept cheap: ubuntu-only, bounds-of-support Pythons (3.10, 3.14) — the full OS/Python matrix already runs in PR CI against the lockfile.

Also documents the workflow in CONTRIBUTING.md ("Tests and checks") and the AGENTS.md quick reference, per the repo's docs-ship-with-the-code convention.

## Validation

Ran the workflow's steps locally: `uv lock --upgrade` + `uv sync --all-extras` + `instro --version` + `just test-python` → 1566 passed (one failure from a local untracked example file that doesn't exist in the repo). The workflow can be exercised in CI via `workflow_dispatch` once merged.


nhschwab:
Note on the pin/revert/re-pin history: the Semgrep findings were first addressed by SHA-pinning, then reverted to match the repo's mutable-tag convention, and now restored since #363 adopts SHA pinning repo-wide (with Dependabot keeping pins current). This branch's workflow now matches that convention.

cartercanedy (approved):

PR #359, opened 2026-08-06
feat(awg): add keysight 33500b awg driver

### LINKED ISSUE #348: driver: add Keysight 33500B AWG driver
### Vendor

Keysight

### Model

33500B

### Category

None of the above / a new category

### Protocol / transport

SCPI over VISA (USB, LAN, GPIB)

### Programming manual or datasheet

https://www.keysight.com/us/en/assets/9018-03290/user-manuals/9018-03290.pdf

### Can you contribute the driver?

Yes — I have the device and plan to write the driver.

### Additional context

_No response_

LizzyC-115:
## Summary

Adding support for a new AWG device. This PR builds on the AWGBaseDriver and InstroAWG framework already in main and implements those functions for the Keysight 33500B series. 

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification
Screenshot 2026-08-06 at 12 48 21 PM
Screenshot 2026-08-06 at 12 48 59 PM

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for Reviewers

- Removal of check_errors was a recent, purposeful change. BLUF: move check_errors inside driver and avoid SCPI specific protocols.
- Modulation commands require the modulation type for all commands on the Keysight. However there is no query for one type of modulation when it is not already enabled. So, we use set_modulation to enable to allow the user to call get_modulation.

### DIFF
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -262,6 +262,11 @@ Optional overrides (raise `NotImplementedError` if not supported):
 - **`get_modulation_type(channel)`**: Read back the active modulation type from the instrument.
 - **`get_modulation_state(channel)`**: Read back whether modulation is enabled from the instrument.
 
+### Modulation
+
+`set_modulation()` only configures modulation on a specific channel and does not enable modulation. To enable or disable, use `modulation_enable()`. 
+`get_modulation_type` returns the most previous modulation type set by the user. 
+
 ### Talking to the Instrument
 
 Concrete drivers should hide transport details behind private attributes. For VISA-backed drivers, create a `VisaDriver` internally and use it for all I/O:
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -264,7 +264,7 @@ Optional overrides (raise `NotImplementedError` if not supported):
 
 ### Modulation
 
-`set_modulation()` only configures modulation on a specific channel and does not enable modulation. 
+`set_modulation()` only configures modulation on a specific channel and does not enable modulation. Only exception is the Keysight 33521B driver: set_modulation will temporarily disable modulation, if previously enabled, to change modulation type. Enable state is restored once `set_modulation` finishes successfully. In error cases, `set_modulation` could leave modulation disabled.
 To enable or disable modulation, use `modulation_enable()`. Only one modulation type is enabled at a time.
 `get_modulation_type` returns the most previous modulation type set by the user. 
 
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -264,7 +264,8 @@ Optional overrides (raise `NotImplementedError` if not supported):
 
 ### Modulation
 
-`set_modulation()` only configures modulation on a specific channel and does not enable modulation. To enable or disable, use `modulation_enable()`. 
+`set_modulation()` only configures modulation on a specific channel and does not enable modulation. 
+To enable or disable modulation, use `modulation_enable()`. Only one modulation type is enabled at a time.
 `get_modulation_type` returns the most previous modulation type set by the user. 
 
 ### Talking to the Instrument
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -16,7 +16,7 @@ description: Using InstroAWG for SCPI-based arbitrary waveform generators
 ## Supported Vendors
 
 - **Rigol**: DG1022Z (DG1000Z series) via SCPI/VISA (`RigolDG1022Z`)
-- **Keysight**: 33521B (33500B series) via SCPI/VISA (`Keysight33500B`)
+- **Keysight**: 33521B (33500B series) via SCPI/VISA (`Keysight33521B`)
 
 If your vendor or model is not listed, see [Custom Driver Development](#custom-driver-development) below.
 
@@ -297,12 +297,6 @@ class RigolDG1022Z(AWGDriverBase):
     def close(self) -> None:
         self._visa.close()
 
-    def check_errors(self) -> None:
-        err = self._visa.query(":SYST:ERR?")
-        code = err.strip().split(",", 1)[0].lstrip("+")
-        if code != "0":
-            raise RuntimeError(f"Rigol DG1022Z reported error: {err.strip()}")
-
     def set_waveform(self, channel: int, waveform: Waveform) -> None:
         with self._visa.lock():
             if isinstance(waveform, Sine):
@@ -312,6 +306,7 @@ class RigolDG1022Z(AWGDriverBase):
             # ... other Waveform shapes ...
             else:
                 raise ValueError(f"unsupported waveform definition {type(waveform).__name__}")
+            self._check_errors()  # one check per call, not one per write — see note above
 
     def set_amplitude(self, channel: int, amplitude: float, unit: AmplitudeMeasurementUnit) -> None:
         if unit is AmplitudeMeasurementUnit.VP:
@@ -319,8 +314,15 @@ class RigolDG1022Z(AWGDriverBase):
         with self._visa.lock():
             self._visa.write(f":SOUR{channel}:VOLT:UNIT {unit.value}")
             self._visa.write(f":SOUR{channel}:VOLT {amplitude}")
+            self._check_errors()
 
     # ... other required and optional methods ...
+
+    def _check_errors(self) -> None:
+        err = self._visa.query(":SYST:ERR?")
+        code = err.strip().split(",", 1)[0].lstrip("+")
+        if code != "0":
+            raise RuntimeError(f"Rigol DG1022Z reported error: {err.strip()}")
 ```
 
 ## Using a Custom Driver
@@ -346,14 +348,10 @@ class MyCustomAWGDriver(AWGDriverBase):
     def close(self) -> None:
         self._visa.close()
 
-    def check_errors(self) -> None:
-        err = self._visa.query("ERR?")
-        if err != "OK":
-            raise RuntimeError(f"AWG error: {err}")
-
     def set_waveform(self, channel: int, waveform: Waveform) -> None:
         if isinstance(waveform, Sine):
             self._visa.write(f"CH{channel}:WAVE SIN,{waveform.frequency_hz}")
+            self._check_errors()
         else:
             raise ValueError(f"unsupported waveform definition {type(waveform).__name__}")
 
@@ -361,6 +359,11 @@ class MyCustomAWGDriver(AWGDriverBase):
 
     def get_output_state(self, channel: int) -> bool:
         return self._visa.query(f"CH{channel}:OUTP?") == "ON"
+
+    def _check_errors(self) -> None:
+        err = self._visa.query("ERR?")
+        if err != "OK":
+            raise RuntimeError(f"AWG error: {err}")
 
 
 awg = InstroAWG(
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -16,6 +16,7 @@ description: Using InstroAWG for SCPI-based arbitrary waveform generators
 ## Supported Vendors
 
 - **Rigol**: DG1022Z (DG1000Z series) via SCPI/VISA (`RigolDG1022Z`)
+- **Keysight**: 33521B (33500B series) via SCPI/VISA (`Keysight33500B`)
 
 If your vendor or model is not listed, see [Custom Driver Development](#custom-driver-development) below.
 
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -228,9 +228,6 @@ def open(self) -> None:
 def close(self) -> None:
     """Close the underlying transport."""
 
-def check_errors(self) -> None:
-    """Drain the instrument error queue; raise if any error is pending."""
-
 def set_waveform(self, channel: int, waveform: Waveform) -> None:
     """Program channel with the waveform definition; raise ValueError if the definition is unsupported."""
 
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -271,6 +271,10 @@ Concrete drivers should hide transport details behind private attributes. For VI
 
 `VisaDriver` owns the resource lock. Concurrent `write` / `query` calls against the same driver are serialized automatically. Use `with self._visa.lock():` when a sequence of writes and their error check need to execute atomically.
 
+Error checking (e.g. SCPI's `SYST:ERR?` queue) is a protocol concept, not an instrument concept, so `AWGDriverBase` has no `check_errors` hook — `InstroAWG` never checks for errors itself. Drivers that talk to an instrument with an explicit error queue own that entirely with a private `_check_errors()`. Drivers on transports with no error-queue concept (most non-SCPI protocols raise immediately on a bad command) simply don't need this at all.
+
+Default to one `_check_errors()` call per public method, after all of that call's writes/queries — cheap, and still catches anything the call did wrong. Only reach for checking after every individual write when checking late is itself a problem: `RigolDG1022Z`'s arbitrary-waveform point-by-point upload (up to 16384 individual writes) checks every point, because the error queue isn't drained and a bad point could otherwise go unnoticed for thousands of writes on a path with a known history of USB/firmware trouble on that hardware. `Keysight33500B`'s arbitrary-waveform download is a single bulk write instead, so it doesn't have that failure mode and sticks with the one-check-per-call default. That's a judgment call each driver author makes for their instrument, not a fixed rule.
+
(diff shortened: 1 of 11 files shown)


packages/instro-unstable/instro/unstable/awg/types.py
```diff
@@ -10,7 +10,7 @@ class AmplitudeMeasurementUnit(Enum):
     VP = "VP"
     VRMS = "VRMS"
     DBM = "DBM"
-
+
```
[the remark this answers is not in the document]
LizzyC-115: This was a nonchanging commit to attempt to re-run the CI during the GIt outage

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,286 @@
...
+                self._visa.write(f"{prefix}:{_MOD_MAGNITUDE_NODE[mod_type]} {magnitude}")
+            elif mod_type is ModulationType.FSK:
+                self._visa.write(f"{prefix}:INT:RATE {modulator.frequency_hz}")
+                self._visa.write(f"{prefix}:FREQ {magnitude}")
+            elif mod_type is ModulationType.PSK:
+                self._visa.write(f"{prefix}:INT:RATE {modulator.frequency_hz}")
+                self._visa.write(f"{prefix}:PHAS {magnitude}")
+            else:
+                raise AssertionError(f"unhandled ModulationType {mod_type}")
+            self._visa.write(f"{prefix}:STAT ON")
```
LizzyC-115: Keysight 33500B require it's SCPI commands to use the modulation type to enable modulation. However, there is no "query for modulation type" SCPI command. The most we can do is check if a modulation type if ON. So, to avoid having software states, choosing to enable modulation in the set_modulation function where the modulation type is unambiguous and force modulation_enable() to disable only.

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33500b.py
```diff
@@ -0,0 +1,286 @@
...
+        _check_channel(channel)
+        if enable:
+            raise ValueError(
+                "the Keysight 33500B enables modulation as part of set_modulation;"
+                " modulation_enable only supports disabling (enable=False)"
+            )
+        with self._visa.lock():
+            for prefix in _MOD_SCPI_PREFIX.values():
+                self._visa.write(f"{prefix}:STAT OFF")
+            self.check_errors()
```
LizzyC-115: Echoing from above function: Keysight 33500B require it's SCPI commands to use the modulation type to enable modulation. However, there is no "query for modulation type" SCPI command. The most we can do is check if a modulation type if ON. So, to avoid having software states, choosing to enable modulation in the set_modulation function where the modulation type is unambiguous and force modulation_enable() to disable only.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -40,10 +40,6 @@ def open(self) -> None:
     def close(self) -> None:
         """Close the underlying transport."""
 
-    @abc.abstractmethod
-    def check_errors(self) -> None:
-        """Check the instrument error queue."""
-
```
LizzyC-115: This file only contains removals of check errors. The reasoning for this change was detailed in this issue here: https://github.com/nominal-io/instro/issues/369

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,327 @@
...
+                    raise ValueError(
+                        f"the Keysight 33500B accepts {_ARB_MIN_POINTS} to {_ARB_MAX_POINTS} arbitrary points"
+                        f" per download, got {num_points}"
+                    )
+                samples_csv = ",".join(str(sample) for sample in waveform.samples)
+                self._visa.write(f"DATA:ARB {_ARB_NAME}, {samples_csv}")
+                self._visa.write(f"FUNC:ARB:SRAT {waveform.sample_rate_hz}")
+                self._visa.write(f"FUNC:ARB {_ARB_NAME}")
+                self._visa.write("FUNC ARB")
+                self._check_errors()
```
LizzyC-115, line 111: Extra check_errors() to raise before setting a software state. Software state is used here because arbitrary waveform samples are not easily queried like the Rigol. So, we need to add extra error checking to make sure state is consistent

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -181,7 +173,6 @@ def _execute_command(
         with self._resource_lock:
             driver_method(channel, value)
             timestamp = time.time_ns()
-            self._check_errors()
         descriptor = f"ch{channel}.{channel_suffix}.cmd"
```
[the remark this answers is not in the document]
LizzyC-115: Rigol driver was modified with the latest awg PR to check_errors in every function, similar to how InstroAWG was structured. The removal of check errors was purposeful to allow for non-SCPI instruments in Instro

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,327 @@
...
+                self._visa.write(f"{prefix}:{_MOD_MAGNITUDE_NODE[mod_type]} {magnitude}")
+            elif mod_type is ModulationType.FSK:
+                self._visa.write(f"{prefix}:INT:RATE {modulator.frequency_hz}")
+                self._visa.write(f"{prefix}:FREQ {magnitude}")
+            elif mod_type is ModulationType.PSK:
+                self._visa.write(f"{prefix}:INT:RATE {modulator.frequency_hz}")
+                self._visa.write(f"{prefix}:PHAS {magnitude}")
+            else:
+                raise AssertionError(f"unhandled ModulationType {mod_type}")
+            self._visa.write(f"{prefix}:STAT ON")
```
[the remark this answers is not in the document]
LizzyC-115: This was a purposeful choice to enable modulation in set_modulation, since Keysight 33521B cannot query for the modulation type but the type is required for enabling. So, to avoid stale software states, we enable in set_modulation and disable in modulation_enable()

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,327 @@
...
+        """Query :SYSTem:ERRor? once and raise on a non-zero code. Does not drain the queue."""
+        err = self._visa.query(":SYST:ERR?")
+        code = err.strip().split(",", 1)[0].lstrip("+")
+        if code != "0":
+            raise RuntimeError(f"Keysight 33500B reported error: {err.strip()}")
+
+
+def _check_channel(channel: int) -> None:
+    if channel != 1:
+        raise ValueError(f"Keysight 33521B only supports 1 channel, got {channel}")
```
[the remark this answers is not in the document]
LizzyC-115: We use the specific model here to tell the user what model this driver is for

packages/instro-unstable/instro/unstable/awg/awg.py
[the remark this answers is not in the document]
LizzyC-115: Let's leave modifying the Rigol driver for a different PR. Drop any issues with the Rigol for now

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
nhschwab: I think the modulation_enable() method should be the only place where a user can turn on/off modulation for a channel. Just because this particular driver doesn't let you query for the modulation type doesn't mean we should break the abstraction. It's the same as the arbitrary waveform download in this driver. Since we can't query the instrument for the arbitrary waveform that we set, we instead cache it in software so that we can return it to the user later. We should do the same for modulation type. If only one modulation type can be enabled at a time, then we can add checks to make sure the hardware and software state do not drift

nhschwab: If a user calls modulation_enable, then we should assume they want to enable the most recently set modulation, so we have to track that in software. We should document this as well

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,336 @@
...
+        _check_channel(channel)
+        with self._visa.lock():
+            if enable:
+                mod_type = self.get_modulation_type(channel)
+                prefix = _MOD_SCPI_PREFIX[mod_type]
+                self._visa.write(f"{prefix}:STAT ON")
+            else:
+                for prefix in _MOD_SCPI_PREFIX.values():
+                    self._visa.write(f"{prefix}:STAT OFF")
+            self._check_errors()
```
LizzyC-115, line 281: We leave the modulation type in `self._last_modulation_type()` even when we disable so the user doesn't have the set the modulation every time they want to do a disable/enable cycle

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,336 @@
...
+                    mod_type
+                    for mod_type, prefix in _MOD_SCPI_PREFIX.items()
+                    if self._visa.query(f"{prefix}:STAT?").strip() == "1"
+                ),
+                None,
+            )
+            self._check_errors()
+            if active is not None and active != self._last_modulation_type:
+                # prioritize hardware state
+                self._last_modulation_type = active
```
LizzyC-115: We defer to the hardware state as truth of which modulation type is enabled. If no modulation type is enabled in hardware, then we use the software state.

LizzyC-115:
Updates hardware and software test runs from updated tests:

Hardware:
Screenshot 2026-08-13 at 12 41 19 PM

Software:
Screenshot 2026-08-13 at 12 41 34 PM

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,324 @@
...
+        if not isinstance(mod_type, ModulationType):
+            raise TypeError(f"mod_type must be a ModulationType, got {type(mod_type).__name__}")
+        if mod_type is ModulationType.ASK:
+            raise ValueError("ASK modulation is not supported by the Keysight 33521B")
+        modulator = _validate_modulator(shape)
+        prefix = _MOD_SCPI_PREFIX[mod_type]
+        with self._visa.lock():
+            was_enabled = self.get_modulation_state(channel)
+            if self._last_modulation_type:
+                self.modulation_enable(channel, False)
```
LizzyC-115: Keysight raises an error when we try to have more than one modulation type enabled at once. This line, coupled with the `was_enabled` logic at the bottom of this function helps to avoid this error. This also helps `set_modulation` persist enable state

LizzyC-115: BLUF: When new modulation type is set, disable all modulation types regardless of state. Then, set the new modulation. Re-enable modulation if modulation was enabled before `set_modulation` was called.

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,324 @@
...
+            # `shape` is the modulator/baseband signal; the carrier is the channel's currently active function.
+            carrier_name = self._visa.query("FUNC?").strip()
+            carrier_type = _FUNC_NAME_TO_CARRIER_TYPE.get(carrier_name)
+            if carrier_type is None:
+                raise ValueError(f"Keysight 33521B reported unsupported waveform '{carrier_name}'")
+            _validate_carrier(mod_type, carrier_type)
```
LizzyC-115, line 238: This line should catch when the user calls set_modulation with the wrong parameters. Address edge case such as setting PWM with non-Pulse waves.

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,324 @@
...
+        with self._visa.lock():
+            if self._last_modulation_type is None:
+                raise RuntimeError(f"channel {channel} has no modulation type currently enabled; set modulation first.")
+        return self._last_modulation_type
+
+    def get_modulation_state(self, channel: int) -> bool:
+        _check_channel(channel)
+        with self._visa.lock():
+            result = any(self._visa.query(f"{prefix}:STAT?").strip() == "1" for prefix in _MOD_SCPI_PREFIX.values())
+            self._check_errors()
```
LizzyC-115, line 297: This should work since in set_modulation, we only have one modulation type on at a given time.

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
nhschwab, line 238: This block should come before the `STAT OFF` writes so that a rejected call is a no-op. As this is currently implemented, a `ValueError` can get raised here after we have already disabled all modulation types and then we are left in an unintended state. We should always try to raise before sending any commands where we can to preserve instrument state prior to the failed method

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,324 @@
...
+        _check_channel(channel)
+        if not isinstance(mod_type, ModulationType):
+            raise TypeError(f"mod_type must be a ModulationType, got {type(mod_type).__name__}")
+        if mod_type is ModulationType.ASK:
+            raise ValueError("ASK modulation is not supported by the Keysight 33521B")
+        modulator = _validate_modulator(shape)
+        prefix = _MOD_SCPI_PREFIX[mod_type]
+        with self._visa.lock():
+            was_enabled = self.get_modulation_state(channel)
+            if self._last_modulation_type:
```
nhschwab: should this gate on `was_enabled` rather than `self._last_modulation_type`?

LizzyC-115: When `_last_modulation_type` is first initialized in `__init__`, it is set to `None`. In this case, none the modulation types have been enabled yet. This gate stops us from having to disable if we have not set modulation type to be anything yet. 

However, if `self._last_modulation_type` is not none, then the user has set a modulation type and we should disable all the modulation types before we set a new type.

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,324 @@
...
+                    self._visa.write(f"{prefix}:STAT OFF")
+            self._check_errors()
+
+    def get_modulation_type(self, channel: int) -> ModulationType:
+        """Returns modulation type currently enabled, or the last type set by the user when none is enabled."""
+        _check_channel(channel)
+        with self._visa.lock():
+            if self._last_modulation_type is None:
+                raise RuntimeError(f"channel {channel} has no modulation type currently enabled; set modulation first.")
+        return self._last_modulation_type
```
nhschwab: Are you trying to protect `self._last_modulation_type` from being modified from another thread? If so, the return state should go inside the lock, otherwise we don't need the lock since the logic is purely in software

nhschwab: Also should the error say "no modulation type currently configured" instead of "enabled'?

packages/instro-unstable/instro/unstable/awg/drivers/keysight_33521b.py
```diff
@@ -0,0 +1,326 @@
...
+                self._visa.write(f"{prefix}:{_MOD_MAGNITUDE_NODE[mod_type]} {magnitude}")
+            elif mod_type is ModulationType.FSK:
+                self._visa.write(f"{prefix}:INT:RATE {modulator.frequency_hz}")
+                self._visa.write(f"{prefix}:FREQ {magnitude}")
+            elif mod_type is ModulationType.PSK:
+                self._visa.write(f"{prefix}:INT:RATE {modulator.frequency_hz}")
+                self._visa.write(f"{prefix}:PHAS {magnitude}")
+            else:
+                raise AssertionError(f"unhandled ModulationType {mod_type}")
+            self._check_errors()
```
nhschwab: let's add a warning log in the event that this raises so that a user knows the modulation set failed and the enabled/disabled state of the modulation may have changed

docs/guides/instrumentation/awg.mdx
```diff
@@ -261,6 +262,12 @@ Optional overrides (raise `NotImplementedError` if not supported):
 - **`get_modulation_type(channel)`**: Read back the active modulation type from the instrument.
 - **`get_modulation_state(channel)`**: Read back whether modulation is enabled from the instrument.
 
+### Modulation
+
+`set_modulation()` only configures modulation on a specific channel and does not enable modulation. 
+To enable or disable modulation, use `modulation_enable()`. Only one modulation type is enabled at a time.
+`get_modulation_type` returns the most previous modulation type set by the user. 
+
```
nhschwab, line 270: Let's document that the keysight driver actually enables/disables modulation in the `set_modulation()` method to be clear to users

nhschwab (approved):

PR #366, opened 2026-08-06
fix(ljm): expose stream buffer byte size configuration

### LINKED ISSUE #360: bug(t8): t8 `hw_timed_high_rate` test failing
Test is failing with
```
 AssertionError: -1483.090101298809 != 5.0 within 0.1 delta (1488.090101298809 difference)
```

When printing out measurements, many are around `5.0` but there's a large chunk that is `-9999.0`. Potential T8 buffer overflow bug.

maxleblang:
## Summary

When a LabJack T8 streams near its max rate (40kHz) there appears to be a firmware bug that causes it to read in a block of `-9999.0` (dropped scan value) values once a stream due to some issue with buffer overflow. This never occurs if the buffer byte size is explicitly set using the `"STREAM_BUFFER_SIZE_BYTES"` register. So we've exposed a way for the user to configure this so they can mitigate these issues if they run into them. By default we set the buffer size to it's default (writing 0 to this register which was the previous behavior).

## Type of change

- [x] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Previously, `test_14_hw_timed_high_rate` for testing the T8 hardware failed due to this issue. Now it, and all other T8 hardware tests, pass.

Screen Shot 2026-08-06 at 05 11 00 869 PM

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

### DIFF
--- a/packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
+++ b/packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
@@ -50,12 +50,23 @@ class LabJackData:
 class LabJackTSeriesDriver(DAQDriverBase):
     """LabJack T-series DAQ driver (T4/T7/T8 via the LJM library)."""
 
-    def __init__(self, device_id: str):
+    def __init__(self, device_id: str, stream_buffer_bytes: int = 0):
+        """Construct the driver.
+
+        Args:
+            device_id: Device serial number, or "ANY" for the first device found.
+            stream_buffer_bytes: Device-side stream buffer size, always written when AI hardware
+                timing is configured so no value carries over from an earlier session.
+                0 selects the device's own default.
+                Must be a power of 2 and the max is 262144.
+                Setting this prevents the device from dropping scans at high sample rates, which LJM reports as -9999 sample values.
+        """
         super().__init__()
         self._model: LJ_Model | None = None
         self._handle: int | None = None
         self._info: tuple[int, int, int, int, int, int] | None = None
         self._device_id = device_id
+        self._stream_buffer_bytes = stream_buffer_bytes
 
         # hw timing settings since LabJack has a single timing engine and samples/channel are predefined
         self._global_scan_rate: float | None = None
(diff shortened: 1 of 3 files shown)


packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -50,12 +50,23 @@ class LabJackData:
 class LabJackTSeriesDriver(DAQDriverBase):
     """LabJack T-series DAQ driver (T4/T7/T8 via the LJM library)."""
 
-    def __init__(self, device_id: str):
+    def __init__(self, device_id: str, stream_buffer_bytes: int = 0):
```
maxleblang, line 53: This is consistent with how `MCCDriver` accepts additional args to its `__init__` method

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series_models.py
```diff
@@ -16,6 +16,7 @@ def hw_timing_configs(
         self,
         hw_timing_config: HWTimingConfig,
         channels: list[AnalogChannel],
+        stream_buffer_bytes: int = 0,
```
maxleblang, line 19: 0 means use the default, which was what we were previously doing

tests/daq/labjack/test_labjack_t8_hardware.py
```diff
@@ -810,7 +815,7 @@ def test_14_hw_timed_high_rate(self):
 
         def step(start_ns: int):
             print(f"         [start {self._ts(start_ns)}]")
-            daq = self._create_daq()
+            daq = self._create_daq(stream_buffer_bytes=HIGH_RATE_STREAM_BUFFER_BYTES)
```
maxleblang, line 818: Removing this arg will cause the test to fail if you want to verify that this is actually fixing the issue

LizzyC-115 (approved):
Additional code only adds extra room to the already existing buffer so existing tests should be enough coverage.

PR #374, opened 2026-08-07
fix(awg): change check_errors() from required to optional in awg

### LINKED ISSUE #369: fix(awg): move check_errors() into each driver, rather than exposing it in InstroAWG
The goal of our AWGDriverBase is to be as generic across AWG instruments as possible. While many AWGs are SCPI/VISA based, this is not true for all devices. To allow for better expansion in the future, we should change the check_errors function in awg.py to be reimplemented inside the device driver itself. This avoids leaking SCPI protocols into the abstraction layer and accommodates non-SCPI/non-VISA instruments at the cost of reimplementing the check_errors() for every SCPI driver.

LizzyC-115:
## Summary

Updating the requirements of the functions in awg. Since we want to allow for broad support of AWGs in the future, we want to avoid using SCPI/VISA specific requirements. `check_errors()` is not universal across non-SCPI instruments. So we make this function optional and simply raise NotImplementedError() and update the documentation to reflect.

## Type of change

- [x] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Screenshot 2026-08-07 at 1 07 52 PM

## Tests

- [ ] Unit tests added or updated
- [x] No tests — explain why: Existing tests were use to verify that this change is non-breaking.

## Checklist

### DIFF
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -227,9 +227,6 @@ def open(self) -> None:
 def close(self) -> None:
     """Close the underlying transport."""
 
-def check_errors(self) -> None:
-    """Drain the instrument error queue; raise if any error is pending."""
-
 def set_waveform(self, channel: int, waveform: Waveform) -> None:
     """Program channel with the waveform definition; raise ValueError if the definition is unsupported."""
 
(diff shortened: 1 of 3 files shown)


maxleblang (approved):
LGTM

hoehner:
I'm thinkin' we move this method _into_ each of the drivers themselves - this method existing on the InstroAWG is leaking VISA / SCPI details into the InstroAWG class - VISA/SCPI details should only ever exist where they are used - in the drivers themselves. This means we actually SHOULD repeat ourselves in each driver.

PR #377, opened 2026-08-07
feat(psu): implement get_voltage_setpoint()/get_current_setpoint() for SiglentSPD3303

### LINKED ISSUE #320: feat(psu): implement get_voltage_setpoint()/get_current_setpoint() for SiglentSPD3303
Part of nominal-io/instro#315.

Implement `get_voltage_setpoint(channel)` and `get_current_setpoint(channel)` on `SiglentSPD3303` (`instro/psu/drivers/siglent_spd3303.py`), reading back the programmed voltage/current setpoint rather than the measured output.

**Not verified yet — no existing evidence in this driver either way.** Unlike the other PSU drivers, `SiglentSPD3303` has no OVP/OCP methods and no other non-measurement (`MEAS:`-prefixed) query used anywhere in the file, so there's no existing precedent confirming a plain `CHn:VOLT?`/`CHn:CURR?` query is even a recognized command on this instrument. This needs real hardware access to confirm the command exists and behaves as expected before assuming the `RigolDP800` pattern applies — check the Siglent SPD3303 programming manual first.

See `RigolDP800`'s implementation (nominal-io/instro#314) as a reference for the general wire-level pattern once confirmed. Needs mocked software tests plus hardware validation against a real unit.

dalvarez-204:
Screenshot 2026-08-12 at 2 03 07 PM
Screenshot 2026-08-12 at 2 04 58 PM
Closes #320.

Implements `get_voltage_setpoint(channel)` / `get_current_setpoint(channel)` on `SiglentSPD3303`, mirroring the `RigolDP800` pattern of reading back the same command node as the `set_*` call, without an argument, rather than the measured value.

## Manual verification

The issue explicitly flagged that this driver has no existing precedent for a non-`MEAS:`-prefixed query, so the command needed confirming against real documentation before writing any hardware-facing code. Found in the Siglent SPD3303X/3303X-E Quick Start manual's SCPI command table:

- Item 7 "VOLTage": `[{CH1|CH2}:]VOLTage?` -- example `CH1:VOLTage?`, typical return `25.000`
- Item 6 "CURRent": `[{CH1|CH2}:]CURRent?` -- example `CH1:CURRent?`, typical return `0.500`

Both are bare floats, same shape as the existing `MEAS:` queries, so no parsing changes were needed -- both new methods go through the existing `_query_checked_float` helper.

## Test plan
- [x] `just check` passes
- [x] Mocked software tests added (`tests/psu/siglent/test_siglent_spd3303_software.py`): happy path for both getters, plus channel-3-unsupported cases
- [x] Hardware test scaffolding added (`tests/psu/siglent/test_siglent_spd3303_hardware.py`), mirroring `RigolDP800`'s setpoint tests -- channel-3/invalid-channel cases too
- [x] Hardware validation against a real SPD3303X/X-E -- passed

## Aside (not fixed here, out of scope)
While reading the manual, noticed its documented `SYSTem:ERRor?` typical return is `0 No Error` (no leading `+`), but this driver's existing `_check_errors` checks for a `+0` prefix. That's pre-existing behavior unrelated to this change -- didn't touch it, but flagging in case it's worth its own ticket (filed as #378).

### DIFF
--- a/instro/psu/drivers/siglent_spd3303.py
+++ b/instro/psu/drivers/siglent_spd3303.py
@@ -33,6 +33,14 @@ class SiglentSPD3303(PSUDriverBase):
         _require_programmable_channel("get_current", channel)
         return self._query_checked_float(f"MEAS:CURR? CH{channel}")
 
+    def get_voltage_setpoint(self, channel: int) -> float:
+        _require_programmable_channel("get_voltage_setpoint", channel)
+        return self._query_checked_float(f"CH{channel}:VOLT?")
+
+    def get_current_setpoint(self, channel: int) -> float:
+        _require_programmable_channel("get_current_setpoint", channel)
+        return self._query_checked_float(f"CH{channel}:CURR?")
+
     def output_enable(self, enable: bool, channel: int) -> None:
         _require_programmable_channel("output_enable", channel)
         cmd = f"OUTP CH{channel},ON" if enable else f"OUTP CH{channel},OFF"
--- a/tests/psu/siglent/test_siglent_spd3303_hardware.py
+++ b/tests/psu/siglent/test_siglent_spd3303_hardware.py
@@ -145,6 +145,24 @@ def test_get_current(driver: SiglentSPD3303, channel_config: ChannelConfig) -> N
 
 
 @pytest.mark.parametrize("channel_config", CHANNELS, ids=lambda config: f"channel_{config.channel}")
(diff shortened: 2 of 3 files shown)


LizzyC-115 (approved):
Utilizes the already approved private helpers to verify channels and uses SPD3303X SCPI commands to set voltage and current. Tests try an invalid and valid command, confirming coverage of potential edge cases.

PR #383, opened 2026-08-10
chore(ljm): add typed channel support for labjacktseriesdriver

maxleblang:
## Summary

This PR implements support for:
- analog voltage input/output channels
- analog thermocouple input channels
- digital input/output channels

for all LabJack T series devices using the new `InstroDAQ` surface. LabJacks do not support current I/O so those are left not implemented and raise a message that points to LabJack's docs about measuring current.

Closes INSTRO-507

## Type of change

- [ ] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Screen Shot 2026-08-10 at 04 26 03 266 PM
Screen Shot 2026-08-10 at 04 29 08 737 PM
Screen Shot 2026-08-10 at 04 32 53 836 PM

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers


packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -47,10 +56,22 @@ class LabJackData:
     dt: int | None
 
 
+def _kelvin_to_unit(temp_k: float, unit: TC_UNIT | None) -> float:
```
maxleblang: All LabJack operations work in units of Kelvin

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -147,10 +178,107 @@ def configure_ai_channel(
...
+        names: list[str] = []
+        for channel in self._ai_channels.values():
+            if isinstance(channel, AnalogThermocoupleChannel) and channel.cjc_source is not CJCSource.CONSTANT:
+                assert self._model is not None
+                name = self._model.tc_cjc_read_name(channel.physical_channel)
+                if name is not None and name not in names:
+                    names.append(name)
+        return names
+
+    def _refresh_tc_cjc(self):
```
maxleblang, line 231: Needed for T4, which needs to snapshot the internal temperature register before streaming (not streamable)

maxleblang:
There's lots of fun LabJack weirdness with configuring thermocouple channels. The T4 flow is widly different than the T7/T8 flow, which between them are also somewhat different. I tried to handle all these nuances while still maintaining a generalized and clean implementation. More info found [here](https://support.labjack.com/docs/using-a-thermocouple-with-labjack)

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
nhschwab: Looks like for both direction temperature unit conversions, we're treating None/default as Celsius? I think it's worth being explicit and adding documentation for devs. Honestly more readable if we have a case for `TC_UNIT.CELSIUS`

maxleblang:
This is the LJM implementation for changes made [here](https://github.com/nominal-io/instro/pull/301) and is work towards #223

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -50,7 +60,7 @@ class LabJackData:
 class LabJackTSeriesDriver(DAQDriverBase):
     """LabJack T-series DAQ driver (T4/T7/T8 via the LJM library)."""
 
-    def __init__(self, device_id: str, stream_buffer_bytes: int = 0):
+    def __init__(self, device_id: str, stream_buffer_bytes: int = 0, tc_input_scaler: Scaler | None = None):
```
nhschwab: I'm very against making the tc_input_scaler a driver-level argument for a few reasons. First off, the amplifiers are per-channel, which means with this approach you assume there can only be one type of amplifier gain. Second, what if a user wants to use an amplifier on a T7 or T8? They can't with this framework. 

I would propose we instead add a field to the `AnalogThermocoupleChannel` that is specifically the input scaler, that way it becomes channel specific and solves the mentioned gaps. The drawback is that it is now shared across DAQ vendors, and NI and MCC handle thermocouple configuration differently, but I think we can ignore/throw a warning in those drivers based on what the struct parameter is set to.

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -147,10 +167,107 @@ def configure_ai_channel(
...
+        else:
+            cjc_k = self._model.tc_cjc_kelvin(channel.physical_channel, cjc_samples)
+
+        tc_type = getattr(ljm.constants, f"tt{channel.tc_type.value}")
+        temps = []
+        for v in volts:
+            try:
+                temps.append(kelvin_to_unit(ljm.tcVoltsToTemp(tc_type, v, cjc_k), channel.unit))
+            except ljm.LJMError:
+                temps.append(float("nan"))  # open/overranged input or dropped-scan sentinel; keep the batch flowing
```
nhschwab: Can we make the exception catch more specific? I feel like there may be some ljm errors that wouldn't necessarily mean what's in the comment

nhschwab: Also can you verify that any downstream handler of measurements do not error with nans? pretty sure this is the right sentinel but worth double checking. Also, we should definitely log a warning or something so the user knows something is wrong thermocouple measurements on a given channel and we should document it clearly. Someone writing logics against a thermocouple channel could get really messed up if a nan sneaks it way in there unexpectedly

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series_models.py
```diff
@@ -125,6 +191,41 @@ def _ai_channel_configs(
...
+        # only write to negative channel if configured channel is even and less than 13
+        if int(channel.physical_channel[3:]) % 2 == 0 and int(channel.physical_channel[3:]) < 13:
+            aNames.append(f"{channel.physical_channel}_NEGATIVE_CH")
+            aValues.append(self._get_negative_channel(None, channel.physical_channel))
+
+        return aNames, aValues
+
+    def tc_cjc_read_name(self, physical_channel: str) -> str:
+        """AIN14 is the internal temp sensor's raw volts; streamable, unlike TEMPERATURE_DEVICE_K."""
+        return "AIN14"
```
nhschwab, line 218: Doesn't labjack let you use any channel as CJC, not just the internal temperature sensor?

maxleblang: For the T4 and T7, LabJack docs talk about using an LM34CAZ CJC for thermocouple reads. However, the configuration between T4 and T7 are extremely different:
- The T4 requires inputting scaling parameters for the specific CJC used that convert the raw voltage to CJC reading
- The T7 requires configuring the given channel as a CJC channel using the `EF` register (which docs call out as non streamable) so we wouldn't be able to stream CJC value, only read it at the start of each stream (T4 behavior)

I don't feel like adding support for channel specific CJC for LabJack is a worthwhile endeavor right now. What are your thoughts?

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
maxleblang: Ya I was having a hard time figuring out a way to fit this in cleanly to the implementation. I think per `AnalogThermocoupleChannel` makes the most sense, I just didn't want it to span across all devices. If you think this is fine I think it's a better implementation.

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
maxleblang: I added a warning and made the error scope smaller.

Using nan is consistent with the rest of the repo as scope also maps an invalid measurement to nan. All downstream publishers also accept nan as well so we're good on that front.

instro/daq/drivers/keysight_34980a.py
```diff
@@ -144,6 +148,12 @@ def configure_ai_channel(
 
         self._ai_channels[channel.alias] = channel
 
+    def configure_ai_thermocouple_channel(self, channel: AnalogThermocoupleChannel):
+        """Thermocouple input is not implemented for Keysight; warns if the LabJack-only ``tc_input_scaler`` is set."""
+        if channel.tc_input_scaler is not None:
+            logger.warning("tc_input_scaler is only honored by the LabJack driver; the Keysight driver ignores it.")
+        super().configure_ai_thermocouple_channel(channel)
```
nhschwab: This feels unnecessary, we already raise an exception that thermocouple input is not configured for the keysight driver, why are we going to warn the user that the keysight driver ignores tc input scaler too?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -252,6 +256,12 @@ def configure_ao_channel(self, channel: AnalogChannel):
 
         self._ao_channels[channel.alias] = channel
 
+    def configure_ai_thermocouple_channel(self, channel: AnalogThermocoupleChannel):
+        """Thermocouple input is not implemented for MCC; warns if the LabJack-only ``tc_input_scaler`` is set."""
+        if channel.tc_input_scaler is not None:
+            logger.warning("tc_input_scaler is only honored by the LabJack driver; the MCC driver ignores it.")
+        super().configure_ai_thermocouple_channel(channel)
```
nhschwab: Same here as keysight. Users can't use thermocouple channels with this driver yet, this warning may confuse them and think there is a path to success when there isn't

packages/instro-daq-ni/instro/daq/drivers/ni/nidaq.py
```diff
@@ -177,6 +181,12 @@ def configure_ao_channel(self, channel: AnalogChannel):
 
         self._ao_channels[channel.alias] = channel
 
+    def configure_ai_thermocouple_channel(self, channel: AnalogThermocoupleChannel):
+        """Thermocouple input is not implemented for NI; warns if the LabJack-only ``tc_input_scaler`` is set."""
+        if channel.tc_input_scaler is not None:
+            logger.warning("tc_input_scaler is only honored by the LabJack driver; the NI driver ignores it.")
+        super().configure_ai_thermocouple_channel(channel)
```
nhschwab: Same here

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series_models.py
nhschwab: I just want to make sure we're on the same page here. Currently, this PR only supports internal and constant CJC configuration for labjack, it does not support channel CJC configuration. I disagree with you that channel CJC is not a worthwhile endeavor. It's common practice to use an external sensor as the CJC, so we will definitely want to support it. I think it's fair to say that it merits its own issue and PR, in which case we should create that issue and mention it in this PR for tracking. For what it's worth I don't think it would be a major lift, just adding CJC channel configuration structs to the daq types module and then making the appropriate insertions in the labjack driver. Up to you if that should be in this PR or not.

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series_models.py
```diff
@@ -125,6 +188,38 @@ def _ai_channel_configs(
...
+
+    def refresh_tc_cjc(self, handle: int | None) -> None:
+        """No-op; CJC is streamable."""
+
+    def tc_cjc_kelvin(self, physical_channel: str, cjc_samples: dict[str, list[float]]) -> float:
+        """AIN14 volts → Kelvin (datasheet §18.0), minus 3 K to reflect screw-terminal temperature."""
+        # CJC moves slowly; the batch's most recent sample is current enough.
```
nhschwab: In the event that a scan is dropped, the most recent value in `cjc_samples` will be -9999.0, which means that in `tc_volts_to_temps` every temperature reading in the `try` block will result in an out of range error and all samples will be dropped as nans and a warning will be logged, even though there are mostly valid samples in that batch. Instead, if we are arguing that CJC moves slowly, we can justify using the most recent sample that is not -9999.0 instead of just the most recent sample to avoid throwing out good samples. Same applies to the T8

docs/guides/instrumentation/daq.mdx
```diff
@@ -264,14 +264,45 @@ Alongside the generic `configure_analog_channel` / `configure_digital_line`, `In
...
+
+```python
+from instro.daq.scaling.thermocouple import TC_TYPE
+
+daq.configure_thermocouple_input("AIN0", TC_TYPE.K, alias="tc0")
+```
+
+Cold-junction compensation depends on `cjc_source`:
+
+- `INTERNAL` (default): the driver reads the device's cold-junction sensor alongside the thermocouple. The T8 streams the per-channel `TEMPERATURE#` sensor, the T7 streams the internal sensor's raw volts (`AIN14`), and the T4 snapshots `TEMPERATURE_DEVICE_K` before each read and conversion (the `TEMPERATURE_DEVICE_K` register cannot be streamed).
```
nhschwab: Might be worth adding a note about hw timed and that T4 CJC may drift over long streaming sessions since it is never refreshed?

nhschwab:
Can we add/update some unit tests for the new conversion helpers and updated measurement read method?

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series_models.py
maxleblang: Ok that makes sense. I think splitting it into it's own issue and linking in this PR makes the most sense as I'll need to order and test with hardware, and I don't want that blocking the merge of this PR. I'll get the hardware ordered so we can add asap.

maxleblang: Issue created: #389 and hardware ordered

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series_models.py
```diff
@@ -125,6 +193,40 @@ def _ai_channel_configs(
...
+        return "AIN14"
+
+    def refresh_tc_cjc(self, handle: int | None) -> None:
+        """No-op; CJC is streamable."""
+
+    def tc_cjc_kelvin(self, physical_channel: str, cjc_samples: dict[str, list[float]]) -> float:
+        """AIN14 volts → Kelvin (datasheet §18.0), minus 3 K to reflect screw-terminal temperature."""
+        samples = cjc_samples[self.tc_cjc_read_name(physical_channel)]
+        # CJC moves slowly; the most recent sample that isn't a dropped-scan sentinel is current enough.
+        volts = next((s for s in reversed(samples) if s != ljm.constants.DUMMY_VALUE), math.nan)
```
maxleblang, line 227: Guard against dropped CJC scans and use the most recent that's valid. If they're all dropped, fallback to `nan` which will propagate up and be handled properly by the driver.

instro/daq/scaling/thermocouple.py
maxleblang, line 1: Changed the TC_UNIT enum (introduced in previous PR so non-breaking) to have naming consistent with pint just in case we go that route. Also handroll kelvin conversions ourselves until we determine a codebase wide way forward for units.

instro/daq/daq.py
nhschwab, line 1: Can you revert this file back to 100644? You can run `chmod 644 instro/daq/daq.py` locally and then push

maxleblang: On it. Need to figure out how to do this on Powershell :(

nhschwab (approved):

PR #387, opened 2026-08-11
feat(dmm): json config-driven instrument creation

nhschwab:
## Summary

Adds config-driven `InstroDMM` initialization from JSON, dict, or a pre-built `DMMConfig`, mirroring the shipped PSU pattern (INSTRO-75 / #131). Implements the schema finalized in [INSTRO-563](https://linear.app/nominal-io/issue/INSTRO-563/define-json-config-schema-for-instrodmm).

Along the way, extracts the config machinery shared across categories into `instro/lib/config.py` and back-ports two hardening fixes to `InstroPSU` so the two config-driven HALs behave identically (see **PSU changes** below).

Closes [INSTRO-74](https://linear.app/nominal-io/issue/INSTRO-74/featdmm-json-config-driven-instrument-creation).

## DMM changes

- **`instro/dmm/config.py` (new)**: `DMM_VENDOR_REGISTRY` (`Agilent34401A`, `Keithley2400`, `Keysight34461A`, `SimulatedDMM`), a DMM `VisaDriverConfig` (no `num_channels`), the `measurement` block (`function`, `digits`, `aperture_nplc`/`aperture_seconds` mutually exclusive, `range: number | "auto" | omitted`), and `DMMConfig` with a validator enforcing that `timing` requires `measurement` — matching `start()`'s existing requirement that a measurement function is set before background polling.
- **`InstroDMM(config=..., autostart=False)`**: same construction contract as `InstroPSU`. The `measurement` block is applied through the public setters on `open()` in the runtime API's required order (`set_measurement_function` → `set_digits` → aperture → `set_range`), once per open (re-applied after `close()`), so applying config publishes the same `.cmd` channels the equivalent manual calls would.
- **Failure-path hardening** (from review): `autostart=True` without a config `measurement` block is rejected up front, before touching the instrument; a failed measurement apply in `open()` closes the driver and rolls back `_measurement_config` so `start()` can't later poll a closed session; autostart failures close config-built publishers so no `NominalCorePublisher` write stream is orphaned (caller-passed publishers stay open since they may be shared).
- **Strict validation**: `version: Literal[1]` and `extra="forbid"` on the shared `DeviceInfo`, so typo'd device fields and future-versioned configs fail at load instead of silently passing.

## Shared config module

- **`instro/lib/config.py` (new)**: the blocks identical across category configs — `TimingConfig`, `NominalCorePublisherConfig`, `FilePublisherConfig`, `PublisherConfigType`, `build_publisher`, and the generic `load_config(config, model_cls)` loader (dict / file path / model instance, deep-copied). `instro/psu/config.py` imports and re-exports them; its public API is unchanged. `resolve_{psu,dmm}_from_config` now return only config-built publishers and the HALs merge them with caller-passed ones.

## PSU changes

- **Autostart failure cleanup** (pre-existing leak): `InstroPSU(config=..., autostart=True)` against an unreachable instrument used to orphan the config-built `NominalCorePublisher` write stream when the constructor raised. PSU now has the same guarded autostart block as DMM: on `open()`/`start()` failure, the driver and config-built publishers are closed before re-raising. Regression-tested.
- **`PSUConfig.version`** tightened to `Literal[1]` and `DeviceInfo` strictness (shared fix above) applies to PSU configs too.
- `InstroPSU._resolve_config` replaced by the shared `load_config` (identical behavior, covered by the existing aliasing/path tests).

## Examples, docs, tests

- `examples/dmm/dmm_config.py` with simulated and Keysight 34461A JSON configs; the simulated config targets the DMM simulator's port 5026 and the script documents starting the simulator.
- `tests/dmm/test_dmm_config.py` (~30 tests): construction paths, all validators, apply-order-on-open against a mock driver, failure-path cleanup/rollback, idempotent re-open, publishers, autostart, registry-drift guards. Plus `tests/dmm/simulated/test_dmm_config_example.py`: runs the shipped simulated example config end-to-end against a real `SimulatedDMMServer` socket, mirroring the PSU example test.
- Docs: "From a JSON Config File" section in `dmm.mdx` (constructor params, measurement block, timing⇒measurement rule, autostart requirement), example page + `docs.json` nav, `::: instro.dmm.config` on the reference page, AGENTS.md bullet documenting the shared config module.

## Testing

- `just check-python` clean (ruff format, mypy, ruff lint)
- `just test-python`: 1604 passed; only failure is from an untracked local file not part of this PR

### Keysight DMM Hardware Config Test
CleanShot 2026-08-12 at 14 22 17@2x

### Rigol PSU Hardware Config Test
CleanShot 2026-08-12 at 14 24 43@2x


instro/dmm/config.py
```diff
@@ -0,0 +1,112 @@
...
+    digits: int | None = Field(default=None, description="Resolution in digits; omit to keep the instrument default.")
+    aperture_nplc: float | None = Field(default=None, description="Integration time in power-line cycles.")
+    aperture_seconds: float | None = Field(default=None, description="Integration time in seconds.")
+    range: float | Literal["auto"] | None = Field(
+        default=None,
+        description="Manual range in the function's units, or 'auto'; omit to keep the instrument default.",
+    )
+
+    @model_validator(mode="after")
+    def _aperture_mutually_exclusive(self) -> MeasurementConfig:
```
nhschwab, line 54: `InstroDMM` drivers can only expose one of `set_aperture_seconds` and `set_aperture_nplc`

instro/dmm/config.py
```diff
@@ -0,0 +1,112 @@
...
+    version: int = 1
+    instrument: Literal["InstroDMM"] = "InstroDMM"
+    device: DeviceInfo
+    driver: VisaDriverConfig
+    measurement: MeasurementConfig | None = None
+    timing: TimingConfig | None = None
+    publishers: list[PublisherConfigType] = Field(default_factory=list)
+
+    @model_validator(mode="after")
+    def _timing_requires_measurement(self) -> DMMConfig:
```
nhschwab, line 89: 'InstroDMM' cannot call `start()` until a measurement function gets set, so if the config has a timing block it must also have a measurement block

instro/dmm/dmm.py
```diff
@@ -203,11 +251,27 @@ def start(self) -> None:
         super().start()
 
     def open(self) -> None:
-        """Open the underlying driver."""
+        """Open the underlying driver and apply any configured measurement state."""
         logger.info("Opening DMM '%s'", self.name)
         self._driver.open()
+        self._apply_measurement_config()
```
nhschwab: if config is provided, on open set the measurement function, set the digits, set the aperture, and set the range (if each respective field is included in the config)

instro/lib/config.py
nhschwab, line 1: Pulling out config dataclasses common to all instruments from InstroPSU

instro/psu/config.py
nhschwab, line 1: Pulling out config dataclasses common to all instruments from InstroPSU

instro/psu/psu.py
```diff
@@ -185,22 +187,14 @@ def __init__(
...
-        with open(Path(config)) as f:
-            return PSUConfig.model_validate(json.load(f))
+            try:
+                self.open()
+                self.start()
+            except Exception:
+                self._driver.close()
+                for publisher in config_publishers:
+                    publisher.close()
+                raise
```
nhschwab, line 197: Fixing a potential leak in PSU autostart found while adding config support for DMM

instro/dmm/config.py
hoehner: Just to clarify, this is a business rule we're enforcing or something across all DMMs?

nhschwab: There is already a rule enforced at the `InstroDMM` level that a driver class can only support one of those aperture configuration methods. As a result, we allow for only one of these fields in the config

hoehner (approved):

PR #399, opened 2026-08-17
refactor(opcua): fix timestamp handling

cartercanedy:
## Summary

Fixes modelling of sample timestamps, where the source timestamp was assumed always available and was silently derived from the server timestamp if it wasn't available. It is now correctly optionally-present and there is no implicit coercion of timestamp sources.

## Type of change

- [ ] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [x] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

## Tests

- [ ] No tests — explain why:

Added regression tests that should signal if timestamp sources are confused or otherwise improperly interpreted.

## Checklist

## Notes for reviewers

### DIFF
--- a/crates/instro-opcua/src/client.rs
+++ b/crates/instro-opcua/src/client.rs
@@ -1013,6 +1013,7 @@ mod subscription_loop_tests {
     use anyhow::Context as _;
     use anyhow::Result;
     use anyhow::anyhow;
+    use futures_util::stream;
     use futures_util::Stream;
     use futures_util::StreamExt as _;
     use tokio::sync::mpsc;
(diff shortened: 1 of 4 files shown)


cartercanedy:
> The PR is not yet safe to merge because the subscription loop can still discard a valid buffered poll sample when the two samples have no timestamp source in common.
> 
> The new filter returns no polled sample whenever neither source timestamps nor server timestamps can be compared, preserving the previously reported data-loss path for notifications without a timestamp matching the poll.


This is fine as long as it's documented

hoehner (approved):

PR #403, opened 2026-08-18
feat(mcc): typed channel support for mcc daqs

maxleblang:
## Summary
This PR adds voltage (existing) I/O, current input, and thermocouple input support to the `MCCDriver`.

This also introduces the `ScanEngine` abstraction to `instro-daq-mcc`. In an effort to avoid writing a driver or specific logic for every MCC DAQ, we can categorize MCC DAQs into three categories:
1. Board has `daq_in_scan` → `DaqInScanEngine` (raw counts, TC support).
2. Board only has `a_in_scan` with SCALEDATA → `ScaledAInScanEngine` (device returns ready-to-use values).
3. Neither → no hardware timing; use polled `read_analog()`.
This helps us write specific hardware timed start and read logic for each hardware configuration without introducing complex and hard to read logic directly to the driver. Additionally, the use of an underlying engine that governs specific logic mirrors what we do for LabJack except the engine is chosen by the driver based on the properties queried from the connected MCC device.

The changes in this PR were tested and validated using a USB-1616HS-4 (`DAQInScanEngine`) and a USB-2404-UI (`ScaledAInScanEngine`)

Closes INSTRO-506

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

All existing hardware tests pass on the USB-1616HS-4
Screen Shot 2026-08-18 at 11 31 59 13 AM

All new typed channel hardware tests pass on the USB-2404-UI
Screen Shot 2026-08-18 at 11 30 40 583 AM

Additionally, hardware-timed and software-timed thermocouple reads work on the USB-1616HS. Validated by generating very simple script configuring a thermocouple channel for an MCC DAQ. (not covered by hardware test suite)
Screen Shot 2026-08-17 at 04 34 39 729 PM
Screen Shot 2026-08-17 at 04 32 54 489 PM

## Tests

- [ ] No tests — explain why:

## Checklist


packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+@dataclass(frozen=True)
+class MCCPortInfo:
+    type: DigitalPortType
+    num_bits: int
+
+
+@dataclass(frozen=True)
+class MCCDeviceInfo:
```
maxleblang, line 32: Addressing what Noah said about how querying MCC info calls underlying functions. Query all important device info before device is started (except for `supported_ranges` as that drives an output).

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+        channel_type_list: list[ChannelType] = []
+        gain_list: list[ULRange] = []
+        alias_list: list[str] = []
+
+        # Analog channels go first: daq_in_scan wants ANALOG entries ahead of other channel types.
+        for channel in channels:
+            if isinstance(channel, AnalogThermocoupleChannel):
+                continue
+            if isinstance(channel, AnalogCurrentChannel):
+                # daq_in_scan boards don't support a current front end
```
maxleblang: MCC doesn't have a `daq_in_scan` supported DAQ that also advertises current inputs.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def start(
+        self,
+        channels: list[AnalogChannelUnion],
+        channel_ranges: dict[str, ULRange],
+        hw_timing_config: HWTimingConfig,
+        buffer_multiplier: int,
+    ) -> None:
+        """Validate the channel set, allocate the scan buffer, and launch the background scan."""
+        # No channel-type pre-validation: DAQICHANTYPE under-reports on some boards, so daq_in_scan is the authority.
```
maxleblang, line 194: Was seeing this on the USB-1616HS. Didn't report that it supported CJC or TC channels but actually does

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+        self.aliases = aliases
+        self.actual_rate = actual_rate / 1000 if high_res else actual_rate
+
+    def convert(self, buffer_snapshot: Array, samples_per_channel: int) -> list[float]:
+        """Snapshot -> user-channel engineering units in scan order, with CJC columns dropped."""
+        to_eng_units = ul.to_eng_units if self._device_info.ai_resolution <= 16 else ul.to_eng_units_32
+
+        temps: list[float] = []
+        if ChannelType.TC in self._channel_type_list:
+            temps = self._convert_tc_counts(buffer_snapshot, samples_per_channel)
```
maxleblang, line 289: Takes in entire buffer but only processes and emits temperature channel data

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+        hw_timing_config: HWTimingConfig,
+        buffer_multiplier: int,
+    ) -> None: ...
+
+    def convert(self, buffer_snapshot: Array, samples_per_channel: int) -> list[float]: ...
+
+    def stop(self) -> None: ...
+
+
+class DaqInScanEngine:
```
maxleblang, line 113: This is the "type" of DAQ we currently support on `main` (minus TC and current support). We added support for `a_in_scan` + SCALEDATA

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+        finally:
+            if self.memhandle:
+                try:
+                    ul.win_buf_free(self.memhandle)
+                except Exception:
+                    pass
+                self.memhandle = 0
+
+
+class ScaledAInScanEngine:
```
maxleblang, line 348: Net new

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -37,12 +49,16 @@
 )
 from instro.lib import Measurement
 
+# The range AiChanType.CURRENT reads through; the UL cannot report it (bounds checked in _get_current_range).
+CURRENT_RANGE = ULRange.BIPPT025AMPS
```
maxleblang, line 56: Bipolar +-25mA. Only current range available and we have to set it in software. Can't be queried from device

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -68,17 +84,17 @@ def __init__(self, device_id: str, buffer_multiplier: int = 2):
...
         self._raw_count_prev: int = 0  # Last raw (unsigned-32) cur_count, for rollover reconstruction
         self._count_offset: int = 0  # Accumulated 2**32 rollovers of cur_count
         self._actual_sample_period: int = 0
         self._timestamper: HWTimestamper | None = None
 
+        self._engine: ScanEngine | None = None
+
         self._ai_channel_ranges: dict[str, ULRange] = {}  # Cache for resolved ULRange per AI channel
         self._ao_channel_ranges: dict[str, ULRange] = {}  # Cache for resolved ULRange per AO channel
+        self._ao_supported_ranges: tuple[ULRange, ...] | None = None  # Probed lazily on first AO configure
```
maxleblang, line 100: We originally searched through the entire `ULRange` enum. Now we record what ranges are specifically available on the connected device.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
                 BoardInfo.ADCHANTYPE,
                 AiChanType.VOLTAGE,
             )
-        except Exception:
-            pass
+        except ULError as e:
+            # devices without configurable channel types raise when calling this configuration
+            if e.errorcode not in (ErrorCode.BADCONFIGITEM, ErrorCode.BADBOARDTYPE):
```
maxleblang, line 205: Ok to pass on these errors because fixed config boards will raise when we try to configure the channel type but still be in a good state.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
-                BoardInfo.RANGE,
-                ul_range,
+                BoardInfo.ADCHANTYPE,
+                AiChanType.TC,
             )
-        except Exception:
-            pass
+        except ULError as e:
+            # dedicated MCC TC devices don't expose this config item and raise when calling this configuration
+            if e.errorcode not in (ErrorCode.BADCONFIGITEM, ErrorCode.BADBOARDTYPE):
```
maxleblang, line 262: Same error handling as above

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+            raise ValueError(
+                f"MCC DAQ applies one board-wide temperature scale, but channels {conflicting} are already "
+                f"configured in a different unit than '{channel.alias}' ({channel.unit.name})."
+            )
 
-        # set channel range
-        ul_range = self._get_range(channel)
-        self._ai_channel_ranges[channel.physical_channel] = ul_range
         try:
             ul.set_config(
```
maxleblang, line 253: Sets channel as temp. DAQs without TC support will never reach this point.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+                AiChanType.TC,
             )
-        except Exception:
-            pass
+        except ULError as e:
+            # dedicated MCC TC devices don't expose this config item and raise when calling this configuration
+            if e.errorcode not in (ErrorCode.BADCONFIGITEM, ErrorCode.BADBOARDTYPE):
+                raise
+
+        ul.set_config(
```
maxleblang, line 265: Set's TC type for temp channel

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+        )
 
-        # set channel terminal config of board
-        # Software-timed does not support per-channel terminal config, each channel configuration will update board-wide terminal config
-        try:
-            ul.a_input_mode(self._board_number, self._get_terminal_config(channel.terminal_config))
-        except Exception:
-            pass
+        # Set the temperature scale hardware-timed scans return; t_in takes the scale per call
+        ul.set_config(
```
maxleblang, line 274: Sets units for the temp channel

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -139,49 +156,21 @@ def _get_terminal_config(
...
-        #         gain_list.append(ULRange.NOTUSED)
-
-        return channel_list, channel_type_list, gain_list
-
     def configure_ai_channel(self, channel: AnalogChannel):
-        """Configure an analog input channel on the MCC DAQ device."""
-        ai_info = self._info.get_ai_info()
-        if not ai_info.is_supported:
+        """Deprecated: use ``configure_ai_voltage_channel``. Configure an analog input channel on the MCC DAQ device."""
+        self.configure_ai_voltage_channel(channel)
```
maxleblang, line 164: Wrap new method

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+        self._ai_channel_ranges[channel.alias] = self._get_current_range(channel)
 
         self._ai_channels[channel.alias] = channel
 
     def configure_ao_channel(self, channel: AnalogChannel):
-        """Configure an analog output channel on the MCC DAQ device."""
-        ao_info = self._info.get_ao_info()
-        if not ao_info.is_supported:
+        """Deprecated: use ``configure_ao_voltage_channel``. Configure an analog output channel on the MCC DAQ device."""
+        self.configure_ao_voltage_channel(channel)
```
maxleblang, line 318: Wrap new method

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+            self._ao_supported_ranges = tuple(AoInfo(self._board_number).supported_ranges)
+
+        # The range is never programmed as config: v_out carries it as an argument on every write.
+        self._ao_channel_ranges[channel.alias] = self._get_range(channel, self._ao_supported_ranges)
 
         self._ao_channels[channel.alias] = channel
 
-    def _get_range(self, channel: AnalogChannel) -> ULRange:
+    def configure_ao_current_channel(self, channel: AnalogCurrentChannel):
+        raise NotImplementedError("MCC DAQ analog outputs are voltage only; current output is not supported.")
```
maxleblang: Some MCC devices do support current output. We didn't have any hardware to test this with so I left it `NotImplemented` and opened and issue #404

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -398,34 +451,37 @@ def get_actual_sample_rate(self) -> float | None:
...
-        ai_resolution = self._info.get_ai_info().resolution
-        channel_list, channel_type_list, gain_list = self._build_channel_lists()
-        if len(channel_list) == 0:
+        ai_resolution = self.get_info().ai_resolution
+        if not self._ai_channels:
             raise ValueError("No analog input channels configured")
-        for ch, ch_type, ch_gain in zip(channel_list, channel_type_list, gain_list):
-            if ai_resolution <= 16:
-                eng_value = ul.v_in(self._board_number, ch, ch_gain)
+        for channel in self._ai_channels.values():
```
maxleblang, line 476: ()sw read) Added logic to read specific channel types (was only voltage before)

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -398,34 +451,37 @@ def get_actual_sample_rate(self) -> float | None:
...
+            elif ai_resolution <= 16:
+                eng_value = ul.v_in(self._board_number, ch, ul_range)
             else:
-                eng_value = ul.v_in_32(self._board_number, ch, ch_gain)
+                eng_value = ul.v_in_32(self._board_number, ch, ul_range)
             data.append(eng_value)
         timestamp = time.time_ns()
 
-        return MCCDAQData(data=data, timestamp=timestamp, dt=None)
+        return MCCDAQData(data=data, timestamp=timestamp, dt=None, aliases=list(self._ai_channels))
```
maxleblang, line 496: Pass in aliases so data is properly labeled in scan order

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -556,40 +600,27 @@ def fetch_analog(self) -> MCCDAQData:
...
-        else:
-            data = [
-                ul.to_eng_units_32(self._board_number, gain_list[i % num_chans], buffer_snapshot[i])
-                for i in range(fetch_size)
-            ]
+        # The engine owns the conversion: snapshot counts -> user-channel engineering units.
+        data = engine.convert(buffer_snapshot, samples_per_channel)
 
-        return MCCDAQData(data=data, timestamp=timestamp, dt=self._actual_sample_period)
+        return MCCDAQData(data=data, timestamp=timestamp, dt=self._actual_sample_period, aliases=list(engine.aliases))
```
maxleblang, line 631: Same as above

tests/daq/mcc/test_mccdaq_hardware.py
maxleblang, line 1: Small nit cleanups

tests/daq/mcc/test_mccdaq_typed_channels.py
maxleblang, line 1: This test is more thorough than NI or LabJack because there are more configuration gotchas and corner cases to test.

tests/daq/ni/test_ni_hardware.py
maxleblang, line 1: Quick nit

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+    def start(
+        self,
+        channels: list[AnalogChannelUnion],
+        channel_ranges: dict[str, ULRange],
+        hw_timing_config: HWTimingConfig,
+        buffer_multiplier: int,
+    ) -> None:
+        """Load the channel/gain queue, allocate the scaled buffer, and launch the background scan."""
+        # a_in_scan reads a contiguous ascending span; the gain queue narrows it and sets per-channel ranges.
+        ordered = sorted(channels, key=lambda channel: int(channel.physical_channel))
```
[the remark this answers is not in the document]
maxleblang: This is incorrect. If a device supports temp and doesn't support daq_in, we will run it as scaled_ain. Running as scaled ain means we configure the channel type as temperature so it will return temperature readings. Verified on USB-2404-UI

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,428 @@
...
+        """Convert a snapshot's raw TC samples to temperatures; get_tc_values only reads UL buffers."""
+        count = len(buffer_snapshot)
+        if self.ctype is c_ushort:
+            scratch = ul.win_buf_alloc(count)
+            copy_to_buf = ul.win_array_to_buf
+        elif self.ctype is c_ulong:
+            scratch = ul.win_buf_alloc_32(count)
+            copy_to_buf = ul.win_array_to_buf_32
+        else:
+            raise NotImplementedError("Thermocouple conversion is not supported for >32-bit scan buffers.")
```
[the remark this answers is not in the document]
maxleblang: Improved the error message. There's currently no MCC hardware that will reach this code path.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -308,79 +407,33 @@ def start(self, **kwargs):
...
-            self._memhandle = 0
-            raise
+
+        engine.start(
+            list(self._ai_channels.values()),
+            self._ai_channel_ranges,
+            hw_timing_config,
+            self._buffer_multiplier,
+        )
+        self._engine = engine
```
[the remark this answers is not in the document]
maxleblang: IMO I don't think this is on the driver to manage and we shouldn't be calling `stop` in start without the user being aware. If the user calls `start()` twice and ends up in a weird state that's on the user. I'll add a guard to check if there's an engine already running and just raise if there is (similar to what we do for NI)

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +185,164 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
 
-        # set channel terminal config of board
-        # Software-timed does not support per-channel terminal config, each channel configuration will update board-wide terminal config
-        try:
-            ul.a_input_mode(self._board_number, self._get_terminal_config(channel.terminal_config))
-        except Exception:
-            pass
+        # Set the temperature scale hardware-timed scans return; t_in takes the scale per call
+        ul.set_config(
+            InfoType.BOARDINFO, self._board_number, int(channel.physical_channel), BoardInfo.TEMPSCALE, temp_scale
```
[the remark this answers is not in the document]
maxleblang: SCALEDATA channels use this config so needed.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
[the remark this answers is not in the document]
maxleblang: Fixed along with previous comment fix

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+        # Each TC entry must immediately follow its CJC (cold-junction sensor) entry, per MCC's convention:
+        tc_channels = sorted(
+            (channel for channel in channels if isinstance(channel, AnalogThermocoupleChannel)),
+            key=lambda channel: int(channel.physical_channel),
+        )
+        for channel in tc_channels:
+            tc = int(channel.physical_channel)
+            # CJC0->TC0, CJC1->TC1/TC2, CJC2->TC3 per bank of 4
+            # Tables: files.digilent.com/manuals/USB-1616HS-4.pdf p.18 and /manuals/USB-2527.pdf p.23.
+            cjc = 3 * (tc // 4) + (0, 1, 1, 2)[tc % 4]
```
nhschwab, line 178: Is this true for all daqin scan engine devices or just the USB-1616HS-4 and USB-2527?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+            self._tc_scratch,
+            0,
+            samples_per_channel,
+            get_temp_scale(self._tc_unit),
+            data_array=self._tc_temps,
+        )
+        # publish NaN for bad scan (this is the only returned error code, everything else raises)
+        if err_code != ErrorCode.NOERRORS:
+            logger.warning("get_tc_values reported OUTOFRANGE (open or overranged thermocouple), returning NaN")
+            return [math.nan if temp == -9999.0 else temp for temp in temps]
```
nhschwab: Where is the -9999.0 coming from? Seems like it's mistakenly borrowed from the ljm sentinel?

From mcculw: https://github.com/mccdaq/mcculw/blob/master/mcculw/ul.py#L4211-L4218. It looks like `ErrorCode.OUTOFRANGE` is the only other return type for `err_code`, so we know it's out of range and can immediately return `math.nan` regardless of the float value in `temps`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+        self._tc_copy_func: Callable = ul.win_array_to_buf
+        self._tc_temps: Array | None = None
+        self.memhandle = 0
+        self.buffer_size = 0
+        self.scan_width = 0
+        self.samples_per_channel = 0
+        self.ctype: type = c_ushort
+        self.copy_func: Callable = ul.win_buf_to_array
+        self.aliases: list[str] = []
+        self.actual_rate = 0.0
```
nhschwab, line 135: Is there any potential for weird behavior or potentially lying to the user by initializing some of these fields as 0?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
maxleblang: Yes true for all I just linked the manual to those two because there's no central documentation on it. The only devices that don't conform are the expansion boards MCC sells which have 8 (I'm pretty sure) CJC channels.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+        self._channel_list = channel_list
+        self._channel_type_list = channel_type_list
+        self._gain_list = gain_list
+        self._tc_unit = next(
+            (channel.unit for channel in channels if isinstance(channel, AnalogThermocoupleChannel)), None
+        )
+        self.scan_width = num_chans
+        self.samples_per_channel = hw_timing_config.samples_per_channel
+        self.aliases = aliases
+        self.actual_rate = actual_rate / 1000 if high_res else actual_rate
```
nhschwab, line 281: Do these field updates belong in the `try` branch so that if the scan start fails we don't update them?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+        self._board_num = board_num
+        self._device_info = device_info
+        self.memhandle = 0
+        self.buffer_size = 0
+        self.scan_width = 0
+        self.samples_per_channel = 0
+        self.ctype: type = c_double
+        self.copy_func: Callable = ul.scaled_win_buf_to_array
+        self.aliases: list[str] = []
+        self.actual_rate = 0.0
```
nhschwab, line 363: Same concern as `DaqInScanEngine`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+                ul.win_buf_free(self.memhandle)
+            except Exception:
+                pass
+            self.memhandle = 0
+            raise
+
+        self.scan_width = num_chans
+        self.samples_per_channel = hw_timing_config.samples_per_channel
+        self.aliases = [channel.alias for channel in ordered]
+        self.actual_rate = actual_rate / 1000 if high_res else actual_rate
```
nhschwab, line 430: Same concern as `DaqInScanEngine`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -139,49 +159,21 @@ def _get_terminal_config(
...
+
+    def configure_ai_voltage_channel(self, channel: AnalogChannel | AnalogVoltageChannel):
+        """Configure a voltage analog input channel on the MCC DAQ device."""
+        info = self.get_info()
+        if not info.ai_supported:
             raise ValueError("Analog input is not supported by this device.")
 
-        if not (channel.physical_channel.isdigit() and int(channel.physical_channel) < ai_info.num_chans):
+        # NUMADCHANS changes when a_input_mode switches SE/DIFF, so read it live rather than from the snapshot.
+        num_chans = ul.get_config(InfoType.BOARDINFO, self._board_number, 0, BoardInfo.NUMADCHANS)
```
nhschwab: Are you saying that a channel configured as differential will result in the high/low channel pairing counting as 1 toward the NUMADCHANS field? So if you configure a channel as SE vs DIFF num_chans can increase or decrease by 1?

If that's the case, then the following check is flawed and should be comparing against the highest possible physical channel number (which should never change on a device regardless of termination config) instead of the number of channels.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +188,165 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+                raise
+
+        # The range is never programmed as config: every read and scan call carries it as an argument.
+        self._ai_channel_ranges[channel.alias] = self._get_range(channel, info.ai_supported_ranges)
+
+        # Set the board-wide input mode; software-timed reads don't support per-channel terminal config
+        input_mode = self._get_terminal_config(channel.terminal_config)
+        if input_mode is not None:
+            try:
+                ul.a_input_mode(self._board_number, input_mode)
```
nhschwab: Okay but this makes it seem like you can only set the input mode for all channels on a board, which would mean that `num_chans` above can only be one of two possible values ... ?

nhschwab: Does the `ul.set_config()` method not allow you to get more granular with the channel termination configs on a board?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +188,165 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+        # daq_in_scan boards have no current front end; reject here so a registered channel can't poison start().
+        if info.daqi_supported:
+            raise ValueError(f"Channel '{channel.alias}': daq_in_scan devices do not support current input channels.")
+
+        # NUMADCHANS changes when a_input_mode switches SE/DIFF, so read it live rather than from the snapshot.
+        num_chans = ul.get_config(InfoType.BOARDINFO, self._board_number, 0, BoardInfo.NUMADCHANS)
+        if not (channel.physical_channel.isdigit() and int(channel.physical_channel) < num_chans):
+            raise ValueError(
+                f"Channel '{channel}' must be in the format '#' where # is an integer less than {num_chans}"
+            )
```
nhschwab: Same concern here as in `configure_ai_voltage_channel()`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +188,165 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
+        # SCALEDATA scans return TC samples in this board-programmed scale; t_in/get_tc_values take it per call.
+        ul.set_config(
+            InfoType.BOARDINFO, self._board_number, int(channel.physical_channel), BoardInfo.TEMPSCALE, temp_scale
+        )
+
+        # Scans need a gain entry per channel; the TC front end runs at the device's most sensitive range.
+        if info.ai_supported_ranges:
+            self._ai_channel_ranges[channel.alias] = min(
+                info.ai_supported_ranges, key=lambda ul_range: ul_range.range_max - ul_range.range_min
+            )
```
nhschwab, line 282: Are we forcing all tc channels to be configured at the minimum range or is that was MCC enforces? Would we ever want to expose range configuration for TC channels to users?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
```diff
@@ -196,67 +188,165 @@ def configure_ai_channel(self, channel: AnalogChannel):
...
-                self._board_number,
-                int(channel.physical_channel),
-                BoardInfo.DACRANGE,
-                ul_range,
-            )
-        except Exception:
-            pass
+        # Probing AO ranges physically drives AO0 to each range's minimum, so it only runs when AO is configured.
+        if self._ao_supported_ranges is None:
+            self._ao_supported_ranges = tuple(AoInfo(self._board_number).supported_ranges)
```
nhschwab, line 336: This is good for now, but if we ever add support for hw-timed analog outputs we will want to make sure that a re-configured channel maintains its configured range and does not fallback to minimum range due to the mcculw `AoInfo` check

nhschwab:
Left a handful of comments. Overall, this is a huge architectural improvement and much needed re-factor. Just some small things to double check against mcculw docs and behavior

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
maxleblang: I believe I saw the -9999 as a sentinel for MCC as well but regardless your comment is a much better approach than checking every value.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
maxleblang: No these are all initialized correctly to 0 and all get set at the proper time. Was there a specific initialization value that you were wondering about?

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
maxleblang: I would say no because if there is an `Exception` we always raise so we'll never update those fields if `daq_in_scan` fails. Additionally I want to keep the code block in the try statement as short as possible.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
maxleblang: Same response as `DaqInScanEngine`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
maxleblang: Same response as `DaqInScanEngine`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
maxleblang: I agree. We'll have to think through stability with analog outputs.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
maxleblang: Got rid of this check. It's behavior and logic was inconsistent across different MCC DAQs and this pattern of channel validation against the hardware in `configure_*` isn't consistent with LJM and NI implementations. Configuring a channel that doesn't exist will throw a "Channel doesn't exist" error in `start()`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
maxleblang: Some boards support configure input mode per channel, while others are only entire board. We should try to configure the channel and if that doesn't work we'll configure the whole board. This could lead to a weird state where the user could configure one channel SE and another DIFF and end up having both be DIFF (because most recent config holds) even though our software state says SE and DIFF. Adding a callout in the docs about this because MCC doesn't add a good way to catch this or track conflicting `terminal_configs`

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
maxleblang: Removing this as well.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
maxleblang: This is kind of a work around for MCC's weird behavior. We have no actual way of configuring the TC channel range on the hardware. That is set on the hardware at +-125mV. We just need to have a range in place for the channel to be valid when passed to `a_load_queue`. We can't query for this range, and some boards don't advertise supporting `BIPPT125VOLTS` so we can't just append that because that will fail to start even if TC at +-125mV is supported. So the workaround is we just say we're measuring at the tightest range (which is always valid) and move on. This value is never read by the hardware. I added a call out for this in the docs.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
nhschwab: No, just wanted to double check!

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
nhschwab: Can we try to add support for the expansion boards as well? Somehow check if a configured tc channel is on the expansion board and then grab the right cjc channel? It's possible our best option is to include a model to property mapping where we manually keep track of these properties per device model. If that's the case, let's open a separate issue and leave the expansion board out of scope and document that we don't yet support tc channels on the expansion boards

numadchans_probe.py
nhschwab, line 1: Assuming this is a mistake push

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
nhschwab: The 5 second timeout in this block means that if a user configures a hw-timed sample rate less than 0.2 Hz (which is now allowed with the new `HIGHRESRATE`), then fetches will raise on a `TimeoutError`. We either need to make the timeout dynamic based on sampling rate or drop support for high res rate

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+        """Return ``(channels, channel_types, gains, aliases)`` for ``ul.daq_in_scan``; aliases exclude CJC entries."""
+        channel_list: list[int | DigitalPortType] = []
+        channel_type_list: list[ChannelType] = []
+        gain_list: list[ULRange] = []
+        alias_list: list[str] = []
+
+        # Analog channels go first: daq_in_scan wants ANALOG entries ahead of other channel types.
+        for channel in channels:
+            if isinstance(channel, AnalogThermocoupleChannel):
+                continue
```
nhschwab, line 161:
```suggestion
        for channel in channels:
            if isinstance(channel, AnalogThermocoupleChannel):
                continue
            if isinstance(channel, AnalogCurrentChannel):
                raise ValueError(f"'{channel.alias}': daq_in_scan does not support current channels.")
```
Not explicitly needed currently since we can't hit this branch but good practice for type checking in case the call sites of this method change

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,446 @@
...
+            if chan_type == ChannelType.TC:
+                data.append(temps[temp_index])
+                temp_index += 1
+            else:
+                data.append(to_eng_units(self._board_num, self._gain_list[i % self.scan_width], buffer_snapshot[i]))
+        return data
+
+    def _convert_tc_counts(self, buffer_snapshot: Array, samples_per_channel: int) -> list[float]:
+        """Convert a snapshot's raw TC samples to temperatures; get_tc_values only reads UL buffers."""
```
nhschwab, line 305:
```suggestion
        """Convert a snapshot's raw TC samples to temperatures; get_tc_values only reads UL buffers."""
        if self._tc_unit is None: 
            raise RuntimeError("TC conversion invoked with no thermocouple channels"))
```

packages/instro-daq-mcc/instro/daq/drivers/mcc/mccdaq.py
maxleblang: I'll change it to be dynamic with `max(5.0, 2 * batch_seconds)` so that we can handle the lower bound.

packages/instro-daq-mcc/instro/daq/drivers/mcc/mcc_engines.py
```diff
@@ -0,0 +1,451 @@
...
+from mcculw.device_info import DaqDeviceInfo
+from mcculw.enums import ChannelType, DigitalPortType, ErrorCode, FunctionType, ScanOptions, TempScale, ULRange
+
+from instro.daq.scaling.thermocouple import TC_UNIT
+from instro.daq.types import (
+    AnalogChannelUnion,
+    AnalogThermocoupleChannel,
+    HWTimingConfig,
+    TerminalConfig,
+)
```
nhschwab, line 20:
```suggestion
from instro.daq.types import (
    AnalogChannelUnion,
    AnalogCurrentChannel,
    AnalogThermocoupleChannel,
    HWTimingConfig,
    TerminalConfig,
)
```

docs/guides/instrumentation/daq.mdx
```diff
@@ -313,6 +313,36 @@ daq.configure_thermocouple_input(
...
+
+```python
+from instro.daq.scaling.thermocouple import TC_TYPE, TC_UNIT
+
+daq.configure_thermocouple_input("0", TC_TYPE.K, alias="tc0", unit=TC_UNIT.CELSIUS)
+```
+
+MCC devices apply one board-wide temperature scale, so every thermocouple channel must use the same `unit`; configuring a channel in a different unit raises `ValueError`. During hardware-timed acquisition an open or overranged thermocouple reads `NaN`, matching the LabJack driver.
+
+Thermocouple channels on MCC expansion boards (the CIO-EXP and EXP-GP families) are not supported. The driver configures thermocouple channels on the base board only.
```
maxleblang, line 328: No expansion board support

nhschwab (approved):

PR #407, opened 2026-08-18
fix(discover): report 2 programmable channels for SiglentSPD3303

### LINKED ISSUE #406: fix(discover): SiglentSPD3303 IDN_MAP reports 3 channels, only 2 are programmable
\`instro/lib/discover.py\`'s \`_IDN_MAP\` entry for the Siglent SPD3303 reports \`num_channels=3\`:

\`\`\`python
("SIGLENT TECHNOLOGIES", "SPD3303"): ("psu", "SiglentSPD3303", 3),
\`\`\`

But the driver itself (\`instro/psu/drivers/siglent_spd3303.py\`) treats channel 3 as unprogrammable -- \`_require_programmable_channel\` explicitly raises \`FeatureNotSupportedError\` for channel 3 on every method, with the reason spelled out in the message: "the SPD3303X/X-E channel 3 output is selectable via front panel switch and does not expose a programming interface."

So a discovered PSU built from this entry (e.g. via \`VisaDriverConfig(num_channels=3, ...)\`) would claim a channel that immediately raises on every call. Should be \`2\`, matching what the driver actually supports over SCPI.

dalvarez-204:
Closes #406.

`_IDN_MAP`'s entry for the Siglent SPD3303 reported `num_channels=3`, but channel 3 can never be a fully usable `InstroPSU` channel regardless of which SPD3303 sub-model (X, X-E, C, S, or D) is actually connected -- confirmed against the real Siglent manuals for each:

- No SCPI command exists to set or query CH3's voltage or current on any variant.
- No SCPI command exists to query CH3's output-enable state on any variant (`OUTPut?` isn't documented for it, and `SYSTem:STATus?`'s bit-field only defines CH1/CH2 bits).

(One correction to the driver's existing comment, surfaced while double-checking this: `OUTPut CH3,ON|OFF` -- the *write* direction of output enable -- actually *is* a documented command on SPD3303X/X-E/C, just not on the legacy S/D. That's a separate, real gap in the driver worth its own ticket, not something this PR touches -- the missing voltage/current control and missing output-status query hold across every variant either way, which is enough on its own to justify `num_channels=2` for discovery purposes.)

Changed to `2`, matching what's actually addressable over SCPI on every variant.

## Test plan
- [x] Added a regression test (`test_scan_recognized_siglent_spd3303_reports_two_programmable_channels`) asserting `scan_visa_resources()` reports 2 channels for a matched SPD3303 IDN response
- [x] `ruff format`/`ruff check`/`mypy` clean
- [x] Full test suite passes (1812 passed; pre-existing, unrelated `ModuleNotFoundError: can` in `tests/unstable/transports/test_can_driver.py` is an environment gap from a stale venv against a recent unrelated dependency addition, not something this PR touches)

### DIFF
--- a/instro/lib/discover.py
+++ b/instro/lib/discover.py
@@ -56,7 +56,7 @@ _IDN_MAP: dict[tuple[str, str], tuple[str, str, int | None]] = {
     ("RIGOL TECHNOLOGIES", "DP821"): ("psu", "RigolDP800", 2),
     ("RIGOL TECHNOLOGIES", "DP831"): ("psu", "RigolDP800", 3),
     ("RIGOL TECHNOLOGIES", "DP832"): ("psu", "RigolDP800", 3),
-    ("SIGLENT TECHNOLOGIES", "SPD3303"): ("psu", "SiglentSPD3303", 3),
+    ("SIGLENT TECHNOLOGIES", "SPD3303"): ("psu", "SiglentSPD3303", 2),
     ("B&K PRECISION", "BK85"): ("eload", "BK85XXB", None),
     ("KEYSIGHT TECHNOLOGIES", "DSOX120"): ("scope", "Keysight1200X", 2),
     ("KEYSIGHT TECHNOLOGIES", "EDUX105"): ("scope", "Keysight1200X", 2),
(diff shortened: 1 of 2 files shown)


LizzyC-115 (approved):
This is updating the channel of the IDN map. The follows the precedent of only listing channels that are programmable. Minimal change and should be non-breaking.

PR #411, opened 2026-08-20
feat(eload): json config-driven instrument creation

nhschwab:
## Summary

Adds `InstroELoad(config=...)` accepting an `ELoadConfig`, dict, or JSON file path, mirroring the shipped PSU (#131) and DMM (#387) pattern, plus the config-driven example.

- The optional `load` block (`mode`, `level`, `curr_limit`, `range`, `slew_rate`) is applied through the public setters on `open()`, so it publishes the same `.cmd` channels the equivalent manual calls would. `mode` is required within the block since level/range can't be set without one.
- **The config never enables the input or shorts the output.** The `load` block pre-arms the setpoint; `output_enable(True)` stays an explicit runtime call.
- `curr_limit` is restricted to CV mode (and requires a `level` to be applied with).
- `timing` is valid without a `load` block: polled measurements (`get_voltage`/`get_current`) work regardless of mode, so standalone timing supports passive monitoring.
- No `num_channels` in the driver block, matching DMM.

`ELOAD_VENDOR_REGISTRY` starts with just `BK85XXB` and expands as ELoad drivers are added.

## Example

- `examples/eload/eload_config.py` + `examples/eload/eload_config_bk85xxb.json` (the exact JSON from INSTRO-566): pre-arms CC/1.5 A from config, then enables the input explicitly at runtime.

## Docs

- Config section in `docs/guides/instrumentation/eload.mdx`, new example page + `docs.json` nav entry, `instro.eload.config` in the reference docs, AGENTS.md shared-config sentence.

## Testing

- `tests/eload/test_eload_config.py`: 28 tests mirroring the DMM config suite (apply order, never-enables-input, rollback on failed apply, reopen semantics, registry-drift guards).
- `ELoadConfig` registered in `tests/test_examples.py` so the example JSON is validated in CI.
- `just check-python` clean; full Python suite passes (1937 passed).

## Hardware Validation

CleanShot 2026-08-21 at 14 26 21

CleanShot 2026-08-21 at 14 27 26


instro/eload/eload.py
```diff
@@ -137,16 +181,38 @@ def _execute_measurement(
...
+        try:
+            self._apply_load_config()
+        except Exception:
+            self._mode = None
+            self._load_config_applied = False
+            self._driver.close()
+            raise
         logger.info("Opened E-Load '%s'", self.name)
 
+    def _apply_load_config(self) -> None:
```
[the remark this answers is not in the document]
nhschwab: Out of scope of this PR, new issue tracking: https://github.com/nominal-io/instro/issues/416

instro/eload/eload.py
```diff
@@ -137,16 +181,38 @@ def _execute_measurement(
...
+        """Apply the config's ``load`` block through the public setters, once per open, in the order the API requires."""
+        if self._config is None or self._config.load is None or self._load_config_applied:
+            return
+        load = self._config.load
+        self.set_mode(load.mode)
```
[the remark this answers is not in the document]
nhschwab: Out of scope of this PR, new issue tracking: https://github.com/nominal-io/instro/issues/417

instro/eload/config.py
```diff
@@ -0,0 +1,115 @@
...
+    config: ELoadConfig,
+) -> tuple[str, ELoadDriverBase, list[Publisher], float | None]:
+    """Resolve a validated ELoadConfig into the ``(name, driver, config_publishers, poll_interval)`` InstroELoad needs."""
+    import importlib
+
+    module_path, class_name = ELOAD_VENDOR_REGISTRY[config.driver.name].rsplit(".", 1)
+    driver_cls = getattr(importlib.import_module(module_path), class_name)
+    driver: ELoadDriverBase = driver_cls(config.driver.visa)  # type: ignore[call-arg]
+
+    config_publishers = [build_publisher(p) for p in config.publishers]
```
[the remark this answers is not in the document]
nhschwab: Low urgency, pre-existing bug establish by PSU and DMM, following pattern for Eload and creating issue to track fix: https://github.com/nominal-io/instro/issues/418

instro/eload/eload.py
```diff
@@ -136,17 +180,43 @@ def _execute_measurement(
...
         logger.info("Opened E-Load '%s'", self.name)
 
-    def close(self):
+    def _apply_load_config(self) -> None:
+        """Apply the config's ``load`` block through the public setters, once per open, in the order the API requires."""
+        if self._config is None or self._config.load is None or self._load_config_applied:
+            return
+        load = self._config.load
+        self.set_mode(load.mode)
+        if load.level is not None:
```
nhschwab:
```suggestion
```

instro/eload/config.py
```diff
@@ -0,0 +1,103 @@
...
+    mode: LoadMode = Field(description="Operating mode: 'CC', 'CV', 'CP', or 'CR'.")
+    level: float | None = Field(
+        default=None,
+        description="Operating level in the mode's units (CC: A, CV: V, CP: W, CR: Ω); omit to keep the instrument default.",
+    )
+    range: float | None = Field(
+        default=None,
+        description="Operating range in the mode's units; omit to keep the instrument default.",
+    )
+    slew_rate: SlewRateConfig | None = Field(default=None, description="Per-edge current slew rate.")
```
nhschwab, line 59: Decided to leave out `curr_limit` from the config even though `InstroEload` accepts it as a parameter in `set_level()`. None of the drivers currently support it, and the existing BK driver silently drops it. I don't actually think this will be a parameter future Eload drivers will likely need, but we can add it to a future config verison if needed

hoehner:
A few nits that don't need any action/commits - just worth mentioning:
- I do think its probably okay to have the instantiation of the InstroEload do some kind of load energization on the SOLE basis that we already have an "autostart" feature. So there is some mental prior-art that makes a user comfy with the idea that "config is ingested/instrument is instantiated -> things are going to start happening". Its not a big deal and I think the "never energizes the load on startup" idea is the safer of the two options, but just stating that I wouldn't reject the config ingestion from doing some kind of loading. 
- Part of me wonders how much this is going to apply to the very next ELoad we support - not a problem for today, problem for another day. The bet is that most if not all of it will apply to the next ELoad. 
- Are there "Multi-channel ELoads" in the world? I assume yes... Not a problem for this PR to solve. 
- You've done a good job evaluating the edge cases here - this is well thought out. 
- Docs look solid as well.

hoehner (approved):

instro/eload/config.py
hoehner: Do we want to ticket this to remove or make more ergonomic?

hoehner: Resolving to unblock merge.

PR #413, opened 2026-08-20
fix(ljm): stop() stops the stream

maxleblang:
## Summary

Oh no! Our current `stop()` implementation for the LabJack driver had `ljm.eStreamStop(self._handle)` commented out which means that calling `stop()` on a LabJack backed `InstroDAQ` doesn't actually stop the stream. This is not right. I don't think we were seeing problems due to this because `start()` also tried to stop an existing stream.

This PR adds better tracking of the `_stream_active` flag and actually stops the device's stream when `stop()` is called.

## Type of change

- [x] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Added a test that starts a device, stops it, verifies it's actually stopped, then starts it again. This was physically validated on T4, T7, and T8 hardware:
Screen Shot 2026-08-20 at 10 50 08 506 AM
Screen Shot 2026-08-20 at 10 51 23 960 AM
Screen Shot 2026-08-20 at 10 52 08 692 AM

## Tests

- [ ] No tests — explain why:

## Checklist

### DIFF
--- a/packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
+++ b/packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
@@ -8,9 +8,10 @@ from dataclasses import dataclass
 from queue import Empty, Queue
 from typing import Mapping
 
+from instro.daq.drivers.labjack.t_series_models import LJ_T4, LJ_T7, LJ_T8, LJ_Model
+
 from instro.daq import DAQDriverBase, HWTimingException
 from instro.daq.drivers import HWTimestamper
-from instro.daq.drivers.labjack.t_series_models import LJ_T4, LJ_T7, LJ_T8, LJ_Model
 from instro.daq.scaling.thermocouple import kelvin_to_unit, unit_to_kelvin
 from instro.daq.types import (
     AnalogChannel,
@@ -132,7 +133,6 @@ class LabJackTSeriesDriver(DAQDriverBase):
 
     def close(self):
         """Disconnect from LabJack device."""
-        self.stop()
         if self._handle is not None:
             ljm.close(self._handle)
             self._handle = None
@@ -376,9 +376,13 @@ class LabJackTSeriesDriver(DAQDriverBase):
         self._actual_sample_rate = actual_scan_rate
         self._actual_sample_period = round(1e9 / actual_scan_rate)
 
-        ljm.setStreamCallback(self._handle, self._stream_callback)
-
         self._streaming_active = True
+        try:
+            ljm.setStreamCallback(self._handle, self._stream_callback)
+        except ljm.LJMError:
+            self._streaming_active = False
+            self._stop_stream()
+            raise
 
     def _stream_callback(self, arg):
         # The lock keeps this read strictly before stop()'s eStreamStop, so it cannot fail.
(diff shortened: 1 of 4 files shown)


packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -364,7 +381,11 @@ def start(self, **kwargs):
         self._streaming_active = True
 
     def _stream_callback(self, arg):
-        response = ljm.eStreamRead(self._handle)
+        # The lock keeps this read strictly before stop()'s eStreamStop, so it cannot fail.
+        with self._stream_lock:
```
maxleblang, line 391: Lock is needed because `_stream_callback` is managed by an LJM spawned thread

maxleblang: Also lock overhead is per batch so isn't slowing down reads

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -391,14 +412,16 @@ def read_analog(
     def fetch_analog(
         self,
     ) -> LabJackData:
+        if not self._streaming_active:
+            raise RuntimeError("No active scan. Call start() before fetch_analog().")
         # Is receiving data from the ljm registered callback.
         try:
             callback_data = self._data_queue.get(timeout=5)
```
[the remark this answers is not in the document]
maxleblang: This is fine and the blocking for 5 seconds before raising `TimeoutError` was previous behavior. This is an improvement on that.

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -364,7 +381,11 @@ def start(self, **kwargs):
         self._streaming_active = True
```
LizzyC-115: Does `ljm.setStreamCallback()` call `self._stream_callback`? If so, there might be a race condition where _stream_callback runs first here when `self._streaming_active = False` and returns prematurely. If not, then this looks good to me since the only time we change `_streaming_active` without a lock is in `start()`.

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
```diff
@@ -102,20 +104,35 @@ def open(self):
...
+        """Discard scans the callback queued before the stream stopped."""
+        while True:
+            try:
+                self._data_queue.get_nowait()
+            except Empty:
+                break
 
     def close(self):
         """Disconnect from LabJack device."""
+        self.stop()
```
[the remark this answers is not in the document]
maxleblang: I'll remove tbh

packages/instro-daq-labjack/instro/daq/drivers/labjack/t_series.py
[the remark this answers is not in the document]
maxleblang: Added lock sync

maxleblang: Good catch. Adding `_streaming_active=True` before stream callback call and wrapping in `try`

LizzyC-115 (approved):

PR #432, opened 2026-08-25
feat(awg): use bulk download for small LAN waveforms

### LINKED ISSUE #331: feat(awg): use bulk download for small LAN download
Right now `RigolDG1022Z` always downloads arb waveforms one point at a
time (`:TRAC:DATA:VALue` in a loop). That was the only way
to avoid a firmware bug where a single bulk `:TRAC:DATA VOLATILE,<floats>`
write hangs the bench unit over USB — happens even with Rigol's own manual
example, verbatim, and needs a power cycle to clear. Per-point avoids the
hang but it's slow, ~2 orders of magnitude slower than a bulk write.

The idea here: over LAN, bulk writes actually work fine as long as the
command stays under ~32KB (the firmware's SCPI parser just can't handle
more than that in one command — unrelated to the documented point-count
limit). So we should use bulk on LAN when the waveform fits, and only fall
back to per-point when it's too big. USB should just keep using per-point
always — bulk isn't salvageable there, that's been thoroughly checked
across both VISA backends, not worth re-opening.

Existing test to extend: `tests/unstable/awg/rigol/test_arb_bulk_download_lan.py`.
Driver lives at `packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py`.

Here are the speed tests and errors from PR #307 

Image

tunglambk:
## Summary

Use the Rigol DG1022Z bulk `:TRAC:DATA` command for arbitrary waveforms sent over TCPIP when the complete command, including the write terminator, is smaller than 32 KiB. USB connections and oversized LAN payloads retain the existing per-point download path.

Document the transport behavior in the AWG guide.

Fixes #331

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

- `uv run --frozen pytest -q tests/unstable/awg/rigol/test_rigol_dg1022z_software.py`: 107 passed
- `uv run --frozen pytest -q`: 2077 passed, 2 skipped, 465 deselected, 1 xfailed
- `uv run --frozen ruff format --check`
- `uv run --frozen ruff check`
- `uv run --frozen mypy`

The mocked tests cover a small TCPIP payload, a USB payload, and an oversized TCPIP payload.

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

The size gate counts encoded command bytes plus the configured write terminator and requires the result to remain strictly below 32 KiB.

### DIFF
--- a/docs/guides/instrumentation/awg.mdx
+++ b/docs/guides/instrumentation/awg.mdx
@@ -54,6 +54,8 @@ InstroAWG supports programming a channel with the following waveforms: `Sine`, `
 Certain features are only available on particular waveforms. Driver's own their implementation of features in InstroAWG. Consult your specific driver for exact support.
 </Note>
 
+`RigolDG1022Z` uses a bulk transfer for arbitrary waveforms over TCPIP when the complete command is smaller than 32 KiB. USB connections and larger TCPIP payloads use per-point transfers because bulk writes can hang the instrument firmware in those cases.
+
 ## Creating an InstroAWG Instance
 
 ```python
--- a/packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
+++ b/packages/instro-unstable/instro/unstable/awg/drivers/rigol_dg1022z.py
@@ -26,6 +26,7 @@ _HIGH_Z_SENTINEL = 9.9e37
 
 _ARB_MIN_POINTS = 9
 _ARB_MAX_POINTS = 16384
+_ARB_BULK_COMMAND_MAX_BYTES = 32 * 1024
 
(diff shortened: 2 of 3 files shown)


LizzyC-115:
Thanks for taking this on, this looks great! Adding hardware validation test runs here for extra confidence.

Existing test_rigol_dg1022z_hardware.py pass:
Screenshot 2026-08-28 at 11 06 19 AM


With PR code:
Screenshot 2026-08-28 at 10 44 21 AM

With original code:
Screenshot 2026-08-28 at 10 49 34 AM

LizzyC-115 (approved):

PR #438, opened 2026-08-26
feat(psu): add get_operating_mode query for CV/CC/off state

wharwood:
## Summary

Adds a method on the InstroPSU class that gets the operating mode: constant voltage, constant current, or off. These are enumerated as CC, CV, or OFF. 

## Type of change

- [ ] Bug fix (`fix`)
- [X] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

I've updated the SimulatedPSU driver and the sim to handle this case. See video below:

https://github.com/user-attachments/assets/cbc90c71-6289-4e8c-895b-3d72ff56f6c8

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

None.

### DIFF
--- a/instro/psu/psu.py
+++ b/instro/psu/psu.py
@@ -14,6 +14,7 @@ from instro.lib.config import load_config
 from instro.lib.instrument import publish_command, publish_measurement
 from instro.lib.publishers import Publisher
 from instro.psu.config import PSUConfig, resolve_psu_from_config
+from instro.psu.types import OperatingMode
 
 logger = logging.getLogger(__name__)
 
@@ -68,6 +69,10 @@ class PSUDriverBase(abc.ABC):
     def get_current_setpoint(self, channel: int) -> float:
         """Query the configured current-limit setpoint (amperes) on `channel`. Output may vary from actual measured current outside of constant current mode."""
         raise NotImplementedError(f"get_current_setpoint is not implemented for {type(self).__name__}")
+
+    def get_operating_mode(self, channel: int) -> OperatingMode:
+        """Query whether `channel` is regulating in constant voltage, constant current, or off."""
+        raise NotImplementedError(f"get_operating_mode is not implemented for {type(self).__name__}")
 
     def set_overvoltage_protection_level(self, voltage: float, channel: int) -> None:
         """Set the overvoltage protection threshold (volts) on `channel`."""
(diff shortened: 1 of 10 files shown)


instro/psu/scpi_sim_server.py
maxleblang: We shouldn't be defining `OperatingMode` in this file anymore. We should import from `types.py` now

instro/psu/psu.py
```diff
@@ -315,8 +320,17 @@ def get_voltage_setpoint(self, channel: int, **kwargs) -> Measurement | None:
             **kwargs,
         )
 
+    @publish_command
+    def get_operating_mode(self, channel: int, **kwargs) -> Command:
```
maxleblang, line 324: I understand why you didn't use `_execute_command` or `_execute_measurement` but is querying a configuration a common pattern we want to add more first class support for. Something like an `_execute_query` method that's shaped like `_execute_measurement` but doesn't try to package measurements?

maxleblang (commented):
Overall looks good. Just added a nit and a comment.

instro/psu/psu.py
wharwood: I'm not sure how common it is, but in this case you can't know the state of the supply unless you know four distinct pieces of information available from the supply: the voltage, current, and their limits. You then need to see which of those actuals isn't at it's limit. That tells you state. Or you could just use this query and not have to worry about that. I'm sure I could come up with other things that require this kind of introspection.

maxleblang: Ok makes sense. This was more of just a cosmetic observation that the rest of the getters and setters wrap a `_execute_command` or `_execute_measurement` method call, but this one does the command calling and packaging manually.

instro/psu/scpi_sim_server.py
wharwood: Disagree. Even though this is packaged with Instro, it's independent of Instro from a dependencies perspective. There should be nothing stopping someone from running this standalone.

instro/psu/psu.py
wharwood: I think I better understand your ask here. The scope interface has a couple of queries that return state. That seems to lend itself toward adding the _execute_query decorator. I've opened #440 for this.

instro/psu/scpi_sim_server.py
maxleblang: So you're saying this `OperatingMode` is more like an internal configuration for the specific simulated device? I can agree with keeping them as two separate things though. I still think it's confusing to have them both named `OperatingMode` so maybe rename this to `SimOperatingMode` or something just to differentiate it.

wharwood: Correct. I think the rename is a fine middle ground.

maxleblang (approved):

PR #445, opened 2026-08-29
feat: publish categorical instrument reads as Measurement

### LINKED ISSUE #440: Consider a shared _execute_query helper for categorical/string state queries published as Command
## Context

Raised by @maxleblang in review on #438 (`get_operating_mode`), on `instro/psu/psu.py`:

> I understand why you didn't use `_execute_command` or `_execute_measurement` but is querying a configuration a common pattern we want to add more first class support for. Something like an `_execute_query` method that's shaped like `_execute_measurement` but doesn't try to package measurements?

## The pattern

`InstroPSU` has two generic helpers: `_execute_command` (setter shape: call the driver with a value, package as `Command`) and `_execute_measurement` (getter shape: call the driver with just a channel, package as `Measurement`). Neither fits a getter whose result is categorical/string rather than numeric, because `_package_measurement` is float-only while `_package_command` already supports `str`.

This isn't hypothetical or PSU-only — the same hand-rolled shape (lock → call driver → timestamp → `_package_command(descriptor, val.value, timestamp, **kwargs)`) already exists **four times** across two categories, all for the identical reason (querying an enum/categorical state, `Measurement` can't carry it):

- `InstroScope.get_acquisition_mode` (`instro/scope/scope.py`)
- `InstroScope.get_acquisition_state`
- `InstroScope.get_trigger_status`
- `InstroPSU.get_operating_mode` (added in #438)

## Proposal to evaluate

A shared `_execute_query` helper, mirroring `_execute_measurement`'s calling shape but ending in `_package_command` instead of `_package_measurement`. Since the need is a consequence of the shared `Command`/`Measurement` type system (not vendor- or category-specific behavior), it likely belongs on the common `Instrument` base class in `instro/lib/instrument.py`, next to `_package_command`/`_package_measurement`, rather than being reinvented per-category.

## Wrinkle to resolve

Scope's three existing methods don't currently go through *any* shared executor, even a Scope-local one — each has slightly different side effects (`get_acquisition_mode`/`set_acquisition_mode` both cache into `self._config.acquisition_mode`; `get_trigger_status` doesn't cache anything; some call `self._check_errors()` inline). A new shared helper needs to either accommodate that variation or only be adopted going forward, leaving Scope's three methods as they are.

## Scope

Not a blocker for #438. Worth a dedicated design pass: is this common enough across categories to earn first-class support, and if so, what's the right shape and where does it live.

wharwood:
## Summary

This started as a review comment on a single PSU getter: `get_operating_mode` hand-packaged a `Command` instead of reusing an existing helper, and the reviewer asked whether querying a device's configuration deserved first-class support the way setting one already did. Digging in turned up a bigger problem than a missing helper. The library had three incompatible answers to "how do I publish a categorical read": `Command` misuse in PSU and three Scope getters (which also meant a read and a write silently shared one channel), total silence in six `InstroAWG` getters that read back an enum and never published it at all, and a hand-built envelope in `EspecGL` working around a real crash in the one shared packaging helper that could have supported this correctly. Rather than add a fourth pattern, this settles the actual rule underneath all of them: direction, not value type, decides `Command` vs. `Measurement`, so a read is always a `Measurement` whatever it holds and a write is always a `Command`. It widens `Measurement` to carry strings safely, teaches the publish path to catch a forgotten `.value` on an enum at the moment it happens rather than downstream, and retrofits all thirteen categorical reads across PSU, Scope, AWG, and EspecGL onto that one rule, which along the way gives AWG's six getters real telemetry for the first time in their history. Closes #440.

## Type of change

- [ ] Bug fix (`fix`)
- [X] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification

Added unit tests.

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers

None.


wharwood:
image
This shows that the PSU operating mode is no longer suffixed with "cmd". It's just another measured value.

instro/daq/daq.py
```diff
@@ -1072,9 +1072,8 @@ def _scale_analog_measurement(self, measurements: list[Measurement]) -> list[Mea
             for ch_name, ch_config in self.ai_channels.items():
```
maxleblang, line 1072: This whole section can be simplified down to
```python
for ch_name, ch_config in self.ai_channels.items():
    if ch_config.scaler:
        key = f"{self.name}.{ch_name}"
        raw_values = cast(list[float], measurement.channel_data[key])
        measurement.channel_data[key] = [ch_config.scaler.scale(val) for val in raw_values]
```
When I was benchmarking `InstroDAQ` performance, `_scale_analog_measurement` was the biggest bottleneck so we should try to keep allocations and copies to a minimum here if we can.

instro/lib/instrument.py
```diff
@@ -47,12 +47,28 @@ def wrapper(self: "Instrument", *args, **kwargs):
...
         items = result if isinstance(result, list) else [result]
+        # Validate every item before publishing any of them: a later item failing a check must
+        # not leave an earlier item already sent to every publisher with no way to undo it.
         for item in items:
             if not isinstance(item, Measurement):
                 raise TypeError(
                     f"@publish_measurement on {func.__qualname__} must return Measurement or list[Measurement], "
                     f"got {type(item).__name__}"
                 )
+            for channel, values in item.channel_data.items():
```
maxleblang, line 58: This guard could be written cleaner as 
```python
for channel, values in item.channel_data.items():
    if not values:
        continue
    if not isinstance(values[0], (int, float, str)):
        raise TypeError(
            f"@publish_measurement on {func.__qualname__} got a non-int/float/str value "
            f"{values[0]!r} ({type(values[0]).__name__}) on channel '{channel}'"
```
This costs the same but checks all channels and is easier to read

instro/lib/instrument.py
```diff
@@ -175,14 +191,18 @@ def _package_command(self, channel: str, data: float | bool | str, timestamp: in
...
 
-    def _package_measurement(self, channel: str, data: float | bool, timestamp: int, **kwargs) -> Measurement:
+    def _package_measurement(self, channel: str, data: float | bool | str, timestamp: int, **kwargs) -> Measurement:
         """Build a single-channel `Measurement` namespaced under this instrument.
 
         The published channel key is ``{self.name}.{channel}``. The caller writes the
         full descriptor, so the literal published name appears at the call site.
         """
+        if isinstance(data, str):
+            values: list[float] | list[str] = [data]
```
maxleblang: nit: `values` can't be `list[float]` if we check that `data` is a `str` first. It must be `list[str]`

instro/psu/psu.py
maxleblang, line 236: We should change this to accept `Enum` as well:
```python
with self._resource_lock:
    val = driver_method(channel=channel)
    timestamp = time.time_ns()
val = val.value if isinstance(val, Enum) else val
```

instro/psu/psu.py
```diff
@@ -320,14 +320,14 @@ def get_voltage_setpoint(self, channel: int, **kwargs) -> Measurement | None:
             **kwargs,
         )
 
-    @publish_command
-    def get_operating_mode(self, channel: int, **kwargs) -> Command:
+    @publish_measurement
+    def get_operating_mode(self, channel: int, **kwargs) -> Measurement:
```
maxleblang: Once `_execute_measurement` accepts an `Enum` we can just write this as
```python
def get_operating_mode(self, channel: int, **kwargs) -> Measurement | None:
    """Query whether ``channel`` is regulating in constant voltage, constant current, or off (published as a string)."""
    return self._execute_measurement(
        self._driver.get_operating_mode,
        channel=channel,
        channel_suffix="operating_mode",
        legacy_suffix="operating_mode",
        **kwargs,
    )
```
and have it be consistent with all other getters in the file

packages/instro-unstable/instro/unstable/awg/awg.py
maxleblang, line 1: Same comments as for `psu.py`: We should have `_execute_measurement` accept an Enum and then have the enum getters just call `_execute_measurement` to be consistent with all the methods in this file

instro/lib/publishers/channel_buffer.py
```diff
@@ -246,23 +246,39 @@ def size_bytes(self) -> int:
...
 
 class NumpyInMemoryPublisher(ChannelBufferPublisher):
+    """String-valued channels are stored in a plain deque; a float32 ring cannot hold them."""
+
     def __init__(self, maxlen: int, value_dtype: Any = np.float32):
         super().__init__(maxlen)
         self._value_dtype = value_dtype
 
     def _ensure_channel(self, channel_name: str) -> None:
-        if channel_name not in self._values:
```
maxleblang, line 254: IMO the logic you added to `_extend_values` should live here. Then we just call `_ensure_channel` when the first value arrives. I don't think it should be created lazily in `_extend_values`

maxleblang (changes requested):
Simplification and clean-up

instro/daq/daq.py
wharwood: Applied as suggested. No reads straight from measurement.channel_data now, no more _get_channel() allocation.

instro/lib/instrument.py
wharwood: Concur. Applied as suggested. Still only checks 1 channel (O(1) instead of O(n)), so can miss if something is wrongly typed that isn't at index 0.

instro/psu/psu.py
wharwood: Applied as suggested.

instro/lib/publishers/channel_buffer.py
wharwood: Applied as suggested.

docs/guides/instrumentation/awg.mdx
```diff
@@ -186,9 +186,15 @@ Every measurement/command call produces a channel keyed under `{name}.{descripto
 | `get_sweep_start_hold_time(channel=N)` | `ch{N}.sweep_start_hold_time` | telemetry |
 | `get_sweep_stop_hold_time(channel=N)` | `ch{N}.sweep_stop_hold_time` | telemetry |
 | `get_sweep_return_time(channel=N)` | `ch{N}.sweep_return_time` | telemetry |
+| `get_modulation_type(channel=N)` | `ch{N}.modulation_type` | telemetry |
```
maxleblang, line 189: For the docs here do we want to specify that it returns `telemetry (Enum)` or some indication that it's enum data?

docs/guides/instrumentation/psu.mdx
```diff
@@ -327,7 +327,7 @@ Every measurement/command call produces a channel keyed under `{name}.{descripto
 | `get_output_status(channel=N)` | `ch{N}.enabled` | telemetry |
 | `get_voltage_setpoint(channel=N)` | `ch{N}.voltage.setpoint` | telemetry |
 | `get_current_setpoint(channel=N)` | `ch{N}.current.setpoint` | telemetry |
-| `get_operating_mode(channel=N)` | `ch{N}.operating_mode.cmd` | command |
+| `get_operating_mode(channel=N)` | `ch{N}.operating_mode` | telemetry |
```
maxleblang, line 330: Same as above

instro/lib/instrument.py
```diff
@@ -47,12 +47,26 @@ def wrapper(self: "Instrument", *args, **kwargs):
...
                 raise TypeError(
                     f"@publish_measurement on {func.__qualname__} must return Measurement or list[Measurement], "
                     f"got {type(item).__name__}"
                 )
+            for channel, values in item.channel_data.items():
+                if not values:
+                    continue
+                # Checking only index 0 costs the same as checking `len(values) == 1` (both are
+                # O(1)), but also catches a forgotten `.value` on an Enum at the head of a bulk
+                # channel (waveforms, DAQ samples) instead of skipping bulk channels outright.
```
maxleblang: Comment should be simplified and sound less AI

packages/instro-unstable/instro/unstable/awg/awg.py
maxleblang, line 1: The function signature changes in this file are breaking. I know this is still unstable so not blocking just flagging here.

maxleblang (dismissed):
Just some nits and assuming you've rerun the python tests to verify. LGTM

packages/instro-unstable/instro/unstable/awg/awg.py
wharwood: Acknowledged without change.

docs/guides/instrumentation/awg.mdx
wharwood: It's actually not enum data (and can never be as we raise if that happens). The only addition here would be the addition of strings to floats and bools. I'm going to not change this and keep things consistent across this and the parallel command notation.

instro/lib/instrument.py
wharwood: Corrected.

maxleblang (approved):
LGMT

PR #448, opened 2026-08-31
feat(awg): add json config for InstroAWG

LizzyC-115:
## Summary

This PR starts adding JSON configuration for InstroAWG following the schema defined in https://linear.app/nominal-io/issue/INSTRO-598/define-json-config-schema-for-instroawg. This is PR 1 of a PR stack. This PR implements the config for the MVP features. Extra features (burst, modulation, sweep) will be added in https://github.com/nominal-io/instro/pull/450.

## Type of change

- [ ] Bug fix (`fix`)
- [x] New feature (`feat`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)

## Verification
Hardware tests:
Screenshot 2026-08-31 at 11 44 18 AM

Software tests:
Screenshot 2026-08-31 at 11 44 38 AM

## Tests

- [ ] No tests — explain why:

## Checklist

## Notes for reviewers


packages/instro-unstable/instro/unstable/awg/config.py
```diff
@@ -0,0 +1,268 @@
...
+) -> tuple[str, AWGDriverBase, int, list[Publisher], float | None]:
+    """Resolve a validated AWGConfig into the ``(name, driver, num_channels, config_publishers, poll_interval)`` InstroAWG needs."""
+    import importlib
+
+    module_path, class_name = AWG_VENDOR_REGISTRY[config.driver.name].rsplit(".", 1)
+    driver_cls = getattr(importlib.import_module(module_path), class_name)
+    driver: AWGDriverBase = driver_cls(config.driver.visa)
+
+    config_publishers = [build_publisher(p) for p in config.publishers]
+    poll_interval = config.timing.poll_interval if config.timing is not None else None
```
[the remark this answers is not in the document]
LizzyC-115: Logic in AWG drivers will raise if an invalid channel is used with a descriptive error message. Leaving this issue as is since it will be caught in channel configuration when `set_waveform` is called.

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -346,16 +390,36 @@ def start(self) -> None:
...
+            if channel_config.offset is not None:
+                self.set_offset(channel, channel_config.offset)
+        self._channel_config_applied = True
+
     def close(self) -> None:
         """Close the underlying driver."""
         logger.info("Closing AWG '%s'", self.name)
         super().close()
         self._driver.close()
+        self._channel_config_applied = False
```
nhschwab:
```suggestion
        self._channel_config_applied = False
        try:
            super().close()
        finally:
            self._driver.close()
```

Prevents a publisher close failure leaking

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -346,16 +390,36 @@ def start(self) -> None:
...
+        """Apply the config's ``channels`` block through the public setters, once per open."""
+        if self._config is None or self._channel_config_applied:
+            return
+        for channel_key, channel_config in self._config.channels.items():
+            channel = int(channel_key)
+            self.set_waveform(channel, channel_config.waveform_definition)
+            if channel_config.amplitude is not None:
+                self.set_amplitude(channel, channel_config.amplitude.value, channel_config.amplitude.unit)
+            if channel_config.offset is not None:
+                self.set_offset(channel, channel_config.offset)
```
nhschwab: Is it possible that for a static waveform a configured offset silently replaces the configured DC level for the channel? How should we handle this?

packages/instro-unstable/instro/unstable/awg/awg.py
```diff
@@ -346,16 +390,36 @@ def start(self) -> None:
...
 
     def open(self) -> None:
-        """Open the underlying driver."""
+        """Open the underlying driver and apply any configured channel state."""
         logger.info("Opening AWG '%s'", self.name)
         self._driver.open()
+        try:
+            self._apply_channel_config()
+        except Exception:
+            self._channel_config_applied = False
```
nhschwab: This line is dead code because `self._channel_config_applied` is only set to True if the try block succeeds. But we do need to drop the configured keys in `self._channel_waveforms` since the successfully configured channels stay defined despite a failure in the try block.

packages/instro-unstable/instro/unstable/awg/config.py
```diff
@@ -0,0 +1,268 @@
...
+    def _build_waveform_definition(self) -> ChannelConfig:
+        """Build the waveform now so ``types.py``'s shape-parameter bounds reject a bad config here, not mid-``open()``."""
+        self._waveform_definition = build_waveform(self.waveform)
+        return self
+
+    @property
+    def waveform_definition(self) -> Waveform:
+        """The runtime ``Waveform`` this channel describes, built when the config was validated."""
+        assert self._waveform_definition is not None
+        return self._waveform_definition
```
nhschwab: The waveform is mutable so there is nothing stopping anyone from constructing the channel config and then changing it later so that `.waveform` and `.wave_definition` drift. You can either make the dataclass frozen or drop the waveform_definition and build the `Waveform` object when the config is applied

packages/instro-unstable/instro/unstable/awg/awg.py
nhschwab: Might make most sense to add a model validation in the config for static waveform that prevents an offset field, your call though

LizzyC-115: Looks like this is also a issue in the actual drivers: a user sets a Static Value with `set_waveform` and then proceeds to use `set_offset` to modify the waveform, despite Static Value being a frozen dataclass. I think if we enforce this in the config, then we should update the driver code too.

I don't think we should update driver code in this PR, but I also don't think we should have diverging behavior for the config versus manual initialization for AWG. So, making an issue for this fix for a later PR: https://github.com/nominal-io/instro/issues/456

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated to match the suggestion

packages/instro-unstable/instro/unstable/awg/config.py
LizzyC-115: Removing the wave_defintion and building at application. Also added updated to have frozen dataclasses to avoid a user passing in a valid config, changing the config, then applying a non-valid config after validation.

packages/instro-unstable/instro/unstable/awg/awg.py
LizzyC-115: Updated! In the exception we will clear the waveforms and close instead. `open()` is the only function that calls `_apply_channel_config()`, so we the try/except block in `open()` should be enough to catch an error

nhschwab (approved):

Issue #418, opened 2026-08-20
Publisher resource leak when build_publisher raises mid-list in config resolution


nhschwab:
## Summary

From a `claude[bot]` inline review comment on #411 (`instro/eload/config.py:113`, also flagged by Greptile): the config resolvers build publishers with a bare list comprehension:

```python
config_publishers = [build_publisher(p) for p in config.publishers]
```

If `build_publisher` raises on the Nth item, publishers 0..N-1 are already constructed with no cleanup path — the exception bypasses them entirely and they are never closed. This pattern exists verbatim in **all three** config resolvers:

- `instro/psu/config.py`
- `instro/dmm/config.py`
- `instro/eload/config.py`

## Why it matters

Publishers have no `__del__`, no `atexit` hook, and no context-manager safety net, so nothing self-cleans by design:

- **`FilePublisher`** (`jsonl`/`csv`/`avro`) opens a real file handle in `__init__`. In CPython, refcounting closes the abandoned fd promptly (with a `ResourceWarning`), and a stray file is left on disk — annoying but mostly self-healing. Relying on refcounting is an implementation detail, not a contract.
- **`NominalCorePublisher`** is the real leak: its constructor authenticates, resolves the dataset, and calls `self._write_stream._impl.open()` on the Rust-backed stream, standing up a network connection and the Rust core's background machinery. Neither it nor `NominalDatasetStream` defines `__del__`, so an abandoned opened stream plausibly holds its connection and threads until process exit.

The failure scenario is realistic: the most likely mid-list failure *is* `NominalCorePublisher` (network/auth), and a config with a Nominal publisher ordered before another failing publisher leaks the already-open stream. In a long-running process that catches the construction error and retries, leaked streams accumulate. Short-lived scripts are unaffected (OS reclaims everything at exit, and nothing was published yet so no data is at risk).

## Suggested fix (from the review comment)

Replace the comprehension with a loop that closes partial results before re-raising, in all three resolvers:

```python
config_publishers: list[Publisher] = []
try:
    for p in config.publishers:
        config_publishers.append(build_publisher(p))
except Exception:
    for pub in config_publishers:
        pub.close()
    raise
```

## Related

- PR #411 (`feat(eload): json config-driven instrument creation`) — where the pattern was flagged; it was ported faithfully from the earlier PSU (#131) and DMM (#387) config PRs, so this is a pre-existing repo-wide pattern, not new to that PR.

nhschwab:
[...]

PR #460, opened 2026-09-03
feat(scope): json config-driven instrument creation

nhschwab:
Closes INSTRO-578. Implements the schema defined in INSTRO-564.

**Adds config-driven workflow for InstroScope. Includes a rename of the existing ScopeConfig to ScopeState to reserve ScopeConfig for the config driven workflow, matching existing instrument configs.**

## Summary

- Add `instro/scope/config.py` with the Pydantic `ScopeConfig` model, its `ChannelConfig` / `AcquisitionConfig` / `TriggerConfig` blocks, `VisaDriverConfig`, and `SCOPE_VENDOR_REGISTRY` for the three registered drivers. Validators enforce the schema rules: channel keys and `trigger.source` within `num_channels`, `average_count` requires `AVERAGE` mode, no duplicate measurements, and `timing` requires at least one polled measurement.
- Add `InstroScope(config=..., autostart=False)` construction from a `ScopeConfig`, dict, or JSON path, mirroring `InstroPSU` / `InstroDMM`. Direct construction is unchanged.
- Register each channel's `measurements` as background daemon functions at construction. `autostart=True` requires at least one; a manual `start()` with none registered logs a warning rather than raising, since starting the daemon is also how the in-memory channel buffer gets created.
- On `open()`, apply the config through the public setters in the order the hardware needs, then `sync_configuration()`, then log one warning per field the instrument reports differently from the config (`SNAP_REL_TOL` = 1%, `SNAP_ABS_TOL` = 1e-9). A mid-apply failure closes the driver and re-raises, and the next `open()` re-attempts the apply. Reopening without closing does not reapply.
- Rename the tracked-state dataclasses `ScopeConfig` / `ChannelConfig` / `TriggerConfig` to `ScopeState` / `ChannelState` / `TriggerState` so the Pydantic models keep the `<Category>Config` names. `sync_configuration()` now returns `ScopeState`. `InstroScope` has no documented external users yet, so this ships without a major bump (see INSTRO-564 Decision 3).

## Apply order on `open()`

Each step exists because a shipped driver needs it:

1. Per channel: `coupling`, `probe_attenuation`, then `vertical_scale`, `vertical_offset`. The probe factor rescales the channel's probe-referred scale and offset on all three drivers.
2. `trigger`: source, type, slope, level, mode.
3. `run()` if `acquisition.start_acquisition_on_open` is set, so the configured trigger is in place when acquisition starts, and so the next step lands on a running scope.
4. `acquisition`: `average_count` before `mode` (Siglent's mode command carries the count inline; Siglent only applies `ACQW` while acquiring), then `horizontal_scale`.
5. `setup_measurement()` for every configured measurement, so Tektronix has its slots installed before the first poll.
6. `sync_configuration()` and the snapped-value warnings.

## Also in this PR

- `Keysight1200X.get_acquisition_mode` keyed its read-back map on long-form names while the instrument answers with the 4-character short form (`AVER`, `HRES`), so AVERAGE and HIGH_RESOLUTION read back as NORMAL. Pre-existing, surfaced by the new post-open check.
- `InstroScope.open()` / `close()` now log lifecycle at info level like the sibling HALs, and `close()` stops the daemon and publishers before closing the driver (scope was the only HAL doing it the other way round), adopting ELoad's flag-reset-then-`finally` shape.

## Deferred

- Publisher leak when the constructor raises after `resolve_scope_from_config` has built publishers: #418 (INSTRO-588), shared by every instrument.
- Reopen after `close()` with attached publishers: INSTRO-606, shared by every instrument.
- Siglent `TRMD` handling: #494 (INSTRO-626), shared root cause for two driver bugs. `run()` writes `TRMD AUTO`, overwriting a configured `trigger.mode: NORMAL` that the snap check cannot catch since trigger fields are not read back; and `set_acquisition_mode()` is dropped by the instrument while stopped, so `acquisition.mode` without `start_acquisition_on_open` on a non-running Siglent does not take effect (the snap check does surface that one). Both are driver fixes and stay out of this PR.

## Docs and examples

- `docs/guides/instrumentation/oscilloscope.mdx`: new "From a JSON Config File" section, `config` / `autostart` parameters, apply-order and lifecycle updates, method-reference row.
- `docs/reference/src/instruments/scope.md`: include `instro.scope.config`.
- `examples/scope/scope_config.py` + `scope_config_keysight_1200x.json`, with the generated example page and nav entry.
- `AGENTS.md`: note the scope config module and the `*State` naming.

## Test plan

- [x] `just check-python` (ruff format, mypy, ruff check) passes
- [x] `uv run pytest tests packages` passes
- [x] `just check-examples` reports no drift
- [x] `tests/scope/test_scope_config.py` covers validation rules, the hardware apply order, the snapped-value warning and its rounding tolerance, rollback and re-apply on a failed open, the `start()` warning, autostart, and vendor-registry drift
- [x] `tests/scope/test_instro_scope.py` covers the Keysight short-form acquisition-mode read-back
- [ ] Hardware smoke against a Keysight 1200X and a Siglent SDS1000X-E with the example config (not run; no scope attached). The apply-order fixes are wire-ordering changes that mocks cannot verify.


instro/scope/scope.py
```diff
@@ -790,8 +913,24 @@ def load_settings(self, name: str, from_instrument: bool = False, **kwargs) -> C
...
 
         return self._package_command("load_settings.cmd", name, timestamp, **kwargs)
+
+
+def _warn_if_snapped(field: str, requested: object, actual: object) -> None:
+    """Warn when the instrument reports a different value than the config requested for ``field``."""
+    if requested is None:
+        return
+    if isinstance(requested, float) and isinstance(actual, (int, float)):
+        matches = math.isclose(requested, actual, rel_tol=1e-3, abs_tol=1e-9)
```
nhschwab: these tolerance values are hard-coded, we need to justify them and possibly make them visible as global variable constants

instro/scope/scope.py
```diff
@@ -294,61 +307,171 @@ def __init__(
...
+            if trigger.slope is not None:
+                self.set_trigger_slope(trigger.slope)
+            if trigger.level is not None:
+                self.set_trigger_level(trigger.level)
+            if trigger.mode is not None:
+                self.set_trigger_mode(trigger.mode)
+        self.sync_configuration()
+        self._warn_on_snapped_values()
+        if acquisition is not None and acquisition.start_acquisition_on_open:
+            self.run()
```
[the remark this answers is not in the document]
nhschwab: Fixed: `_apply_config` now calls the driver's `setup_measurement` for every configured measurement after the trigger block and before `sync_configuration()` and `run()`. Keysight and Siglent are no-ops but this fixes the potential stale result return by the Tek

instro/scope/scope.py
```diff
@@ -294,61 +307,171 @@ def __init__(
...
+
+        if poll_interval is not None:
+            self.background_interval = poll_interval
+
+        if autostart:
+            if not self._background_methods:
+                raise ValueError(
+                    "autostart=True requires a config with at least one channels.<n>.measurements entry; "
+                    "background polling has nothing to do otherwise."
+                )
```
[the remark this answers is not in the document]
nhschwab: Already tracked by issue #418 and shared by other instruments, deferring to the PR that closes that issue

instro/scope/scope.py
```diff
@@ -294,61 +307,171 @@ def __init__(
...
     def open(self) -> None:
-        """Open the underlying driver."""
+        """Open the underlying driver and apply any configured initial state."""
         self._driver.open()
+        try:
+            self._apply_config()
+        except Exception:
+            self._config_applied = False
+            self._driver.close()
+            raise
```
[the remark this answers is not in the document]
nhschwab: Fixed and also added a try, except block to avoid stale configuration

instro/scope/scope.py
nhschwab: Moved both into public module constants, SNAP_REL_TOL and SNAP_ABS_TOL. Values were justified by looking at precision of supported driver returns

instro/scope/drivers/keysight_1200x.py
nhschwab, line 1: Found a bug in `get_acquisition_mode` returns in the keysight driver, fixes how the return strings map to AcquisitionMode types

instro/scope/types.py
nhschwab, line 1: Renames Config --> State to reserve ScopeConfig

instro/scope/scope.py
```diff
@@ -3,26 +3,42 @@
...
     WaveformData,
 )
 
+logger = logging.getLogger(__name__)
+
+# Tolerances for the post-open snapped-value check (config request vs. sync_configuration() read-back).
+# Read-backs are formatted to as few as 3 significant figures (Siglent returns e.g. "2.00E-01"), so
+# rounding alone can move a value by up to 0.5%; 1% absorbs that with margin, while real snaps are far
+# coarser (1-2-5 scale steps change the value by >= 25%). The absolute floor only matters at or near
+# zero, where float parsing noise (1e-12) would otherwise defeat a relative tolerance.
```
nhschwab:
```suggestion
```

instro/scope/drivers/keysight_1200x.py
maxleblang: Just curious how did you find this bug?

instro/scope/scope.py
```diff
@@ -3,26 +3,38 @@
...
     TriggerMode,
     TriggerSlope,
     TriggerStatus,
     TriggerType,
     WaveformData,
 )
 
+logger = logging.getLogger(__name__)
+
+# Tolerances for the post-open snapped-value check (config request vs. sync_configuration() read-back).
```
maxleblang, line 35: How did we decide on these tolerances?

instro/scope/scope.py
```diff
@@ -294,61 +311,202 @@ def __init__(
...
+        poll_interval: float | None = None
+        resolved_config: ScopeConfig | None = None
+        config_publishers: list[Publisher] = []
+        if config is not None:
+            if driver is not None or num_channels is not None:
+                raise ValueError(
+                    "InstroScope(config=...) cannot be combined with driver/num_channels; "
+                    "use one construction style or the other."
+                )
+            resolved_config = load_config(config, ScopeConfig)
```
maxleblang, line 344: Is there a scenario when we would want to load a config but not resolve scope objects from it? Why are `load_config` and `resolve_scope_from_config` two separate functions?

instro/scope/scope.py
```diff
@@ -294,61 +311,202 @@ def __init__(
...
 
     def close(self) -> None:
-        """Close the underlying driver and stop the daemon."""
-        self._driver.close()
-        super().close()
+        """Stop the daemon and publishers, then close the underlying driver."""
+        logger.info("Closing scope '%s'", self.name)
+        # Reset before teardown: a failing publisher close must not strand the flag,
+        # or a later reopen would silently skip re-applying the configured state.
+        self._config_applied = False
```
maxleblang, line 488: I don't know if this makes sense. If I open a Scope with some configuration, modify it while it's running, then close it I would expect it to maintain the state it was last in if I open it again in that script, not start from the initial configuration.

instro/scope/scope.py
```diff
@@ -294,61 +311,202 @@ def __init__(
...
+                self.set_trigger_level(trigger.level)
+            if trigger.mode is not None:
+                self.set_trigger_mode(trigger.mode)
+        acquisition = self._config.acquisition
+        # Start acquiring before the acquisition block: Siglent only applies acquisition-mode
+        # changes while running, and Tektronix measurement slots only settle against a live acquisition.
+        if acquisition is not None and acquisition.start_acquisition_on_open:
+            self.run()
+        if acquisition is not None:
+            # Count before mode: Siglent's mode command carries the average count inline.
```
maxleblang, line 440: This fixes this bug for applying a configuration, but highlights a bug in the Siglent driver without using config flow:
```python
scope = InstroScope("bench", driver=SiglentSDS1000XE("TCPIP0::..."), num_channels=4)
scope.open()

scope.set_acquisition_mode(AcquisitionMode.AVERAGE)   # writes ACQW AVERAGE,16  <- stale default
scope.set_average_count(64)                            # writes AVGA 64
# Scope now averages 16, not 64. No error.
```
We should have an issue open for this if there already isn't one.

instro/scope/scope.py
```diff
@@ -294,61 +311,202 @@ def __init__(
...
+            if trigger.slope is not None:
+                self.set_trigger_slope(trigger.slope)
+            if trigger.level is not None:
+                self.set_trigger_level(trigger.level)
+            if trigger.mode is not None:
+                self.set_trigger_mode(trigger.mode)
+        acquisition = self._config.acquisition
+        # Start acquiring before the acquisition block: Siglent only applies acquisition-mode
+        # changes while running, and Tektronix measurement slots only settle against a live acquisition.
+        if acquisition is not None and acquisition.start_acquisition_on_open:
```
maxleblang, line 437: If `start_acquisition_on_open` is false but the user still tries to configure `acquisition_mode` for a Siglent device won't that configuration be dropped and never applied because the device isn't running when `set_acquisition_mode` is called?

instro/scope/types.py
maxleblang: Are we worried about this being a breaking change for users? If a user was already importing and using `ChannelConfig`, `ScopeConfig`, or `TriggerConfig` they won't get an import error because those still exist but are now completely different objects

instro/scope/scope.py
```diff
@@ -294,61 +311,202 @@ def __init__(
...
+                self.name,
+            )
+        super().start()
+
+    def _warn_on_snapped_values(self) -> None:
+        """Log one warning per config field the instrument reports differently after ``sync_configuration()``."""
+        assert self._config is not None
+        for channel, requested in sorted(self._config.channels.items()):
+            actual = self._state.channels[channel]
+            for attr in ("vertical_scale", "vertical_offset", "coupling", "probe_attenuation"):
```
maxleblang: The list of attrs to check should be derived not hardcoded. That way if another field is added later we don't have to worry about adding it to this tuple

instro/scope/scope.py
```diff
@@ -294,61 +311,202 @@ def __init__(
...
-        return self._config.channels[channel]
+        if resolved_config is not None:
+            for channel, channel_config in sorted(resolved_config.channels.items()):
+                for measurement_type in channel_config.measurements:
+                    self.add_background_daemon_function(self.measure, measurement_type, channel=channel)
+
+        if poll_interval is not None:
+            self.background_interval = poll_interval
+
+        if autostart:
```
maxleblang, line 381: We should put the autostart check above where the publishers get built (`resolve_scope_from_config`) so we don't have to worry about closing them if there is an issue. This also means modifying the check to not look at `_background_methods` but at `resolved_config.channels.measurements`

maxleblang (changes requested):

instro/scope/drivers/keysight_1200x.py
nhschwab: Flagged in an AI review and corroborated by the programming manual. This new approach is robust to both short form and long form returns. See page 185 in the programming manual: http://www.sophphx.caltech.edu/Lab_Equipment/Keysight_1000_X-Series_prog_guide.pdf

Image

instro/scope/scope.py
nhschwab: Good question.

SNAP_REL_TOL = 1e-2 is driven by read-back formatting. The Siglent driver returns scale and offset at three significant figures (e.g. "2.00E-01") while formatting its own writes at five, so a request like 0.1236 V/div goes out as 1.2360E-01 and comes back as 1.24E-01. That's a 0.3% delta from rounding alone. Worst case for three sig figs is half a unit in the last place, or 0.5%, so 1% covers it with 2x margin. Real snaps are much coarser than that since scale steps follow a 1-2-5 sequence, so the smallest genuine snap moves the value by at least 25%. That leaves more than an order of magnitude between rounding noise and anything we'd actually want to warn about.

SNAP_ABS_TOL = 1e-9 exists because a relative tolerance is meaningless at zero. If a requested offset is 0.0, it won't warn on floating noise. A nanovolt or nanosecond is below any scope's resolution, so it can't hide a real change, and it only comes into play for values at or near zero.

Enums and ints (coupling, acquisition mode, average count) still compare exactly, and trigger fields aren't compared at all since sync_configuration() doesn't read them back.

instro/scope/scope.py
nhschwab: `load_config` is a library-wide helper for resolving configs from dicts, paths to jsons, and pre-existing models. `resolve_scope_from_config` is category specific. Keeping them separate means that validation fails fast beofre any instrument is constructed, and there are real use cases for wanting the validated model before construction (CI/testing, automated workflows, runtime edits to the config, etc.)

instro/scope/types.py
nhschwab: Yup, this is definitely a breaking change and good to flag that. After discussing with @hoehner , we decided that since InstroScope is rather new and doesn't have a ton of usage yet, the breaking change risk is acceptable given the importance of standardizing the module naming.

instro/scope/scope.py
nhschwab: Yeah, this is a problem with the device driver though, no the config workflow. On a stopped Siglent the config is dropped but `sync_configuration()` reports the actual acquisition state so it gets reported to the user. If the Siglent is free running when the config is applied then it applies fine. The root of the issue is that the Siglent driver conflates trigger sweep and acquisition state in a single command. Adding a doc sentence for visibility and also opened a tracking issue: https://github.com/nominal-io/instro/issues/494

instro/scope/scope.py
nhschwab: https://github.com/nominal-io/instro/issues/494

instro/scope/scope.py
nhschwab: Agreed for the channel fields. ChannelConfig (pydantic) and ChannelState (dataclass) share names for everything that's comparable, so the tuple can be the intersection of the two:

```
_CHANNEL_SNAP_FIELDS = tuple(
    f.name for f in dataclasses.fields(ChannelState) if f.name in ChannelConfig.model_fields
)
```

measurements falls out automatically since it's config-only, and a field added to both models is picked up. The acquisition fields are excluded since the naming doesn't line up, so we'll still have to call those out manually.

instro/scope/scope.py
nhschwab: I would disagree. The instrument itself is constructed from a config, with that expectation that when that instrument is opened, the config is applied. The workflow you're describing (open instrument, apply commands and update instrument state, close instrument, reopen instrument and have runtime config persist) makes more sense with the direct construction call. I see what you mean though, so maybe we can add a boolean field for reapplying the config in a separate PR?

maxleblang: Ya maybe we could have an `scope.open(open_from_config_state=True)` flag or something. This will establish a precedent for the config flow for all of our instruments though so I think we need to be sure about wanting it done this way.

instro/scope/scope.py
maxleblang: So we would need to ensure a field we want to check for snapping has the exact same name on `ChannelConfig` as it does on `ChannelState`?

maxleblang (dismissed):

instro/scope/scope.py
nhschwab: Yeah the intersection is by name so a field must be spelled the same on both to be compared. The upside is that a config-only or state-only field is excluded automatically. The downside is that a rename on one side drops out silently. I added a drift test asserting set(ChannelConfig.model_fields) - {"measurements"} == set(_CHANNEL_SNAP_FIELDS) to make sure that gets flagged in CI

maxleblang (approved):

Issue #476, opened 2026-09-04
release PR #470 bumps instro and instro-unstable to 2.0.0


hoehner:
**Symptom:** release-please PR #470 proposes instro 2.0.0 and instro-unstable 2.0.0.

**Cause:** #453 landed as `feat(awg)!:`. The `!` marks a breaking change, and both packages are past 1.0, so release-please bumps the major version of every component the commit touched.

**Fix:** land `Release-As` overrides so the release is instro 1.18.0 and instro-unstable 1.11.0. One commit scoped to `packages/instro-unstable` (`Release-As: 1.11.0`), then one scoped to the root package (`Release-As: 1.18.0`). The root override must merge last: the root component sees every commit and the newest `Release-As` wins.

PR #478, opened 2026-09-04
docs(agents): document Release-As scoping for release-please overrides

hoehner:
## Summary

Second of two `Release-As` commits for #476. Carries `Release-As: 1.18.0` for the `instro` component so release PR #470 ships instro 1.18.0 instead of 2.0.0. Documents the scoping rules in the Release PRs section of AGENTS.md.

**Merge order matters:** merge after the instro-unstable override PR, and squash with the footer in the commit body:

```
gh pr merge --squash --body "Release-As: 1.18.0"
```

## Type of change

- [x] Documentation (`docs`)

## Verification

Docs-only change. Release-please mechanics verified against `manifest.ts`, `commit-split.ts`, `commit-exclude.ts`, and `strategies/base.ts` upstream.

## Tests

- [x] No tests — docs only

## Checklist

- [x] PR title follows Conventional Commits
- [x] Documentation updated


hoehner:
Addressed in cbe8da1:

- **crates/ exclusion**: reworded; root `instro` sees every commit except ones whose files all sit under `crates/`.
- **Scoping vs root**: now states explicitly that a sub-package override also reaches root `instro` and cannot be scoped away, hence the merge-last rule.
- **`--body` replaces the generated body**: correct, and now stated. For a dedicated override commit that is the intent; the changelog entry comes from the PR title, which `--body` does not touch.
- **CONTRIBUTING.md**: mirrored a short note in its Release PRs section.

Not changed: "a commit that touches no files is applied to every component" is accurate. `manifest.ts` constructs `CommitSplit` with `includeEmpty: true`, and `CommitSplit.split()` pushes a commit with `files.length === 0` onto every configured package path. Root gets the unsplit list anyway, so an empty commit reaches all components.

Merge with `gh pr merge 478 --squash --body "Release-As: 1.18.0"`.

PR #526, opened 2026-09-14
feat(opcua): include node type references

cartercanedy:
## Stack
1. [#540](https://github.com/nominal-io/instro/pull/540) — recurse all node classes
2. [#541](https://github.com/nominal-io/instro/pull/541) — add OpcUaAttributeId
3. [#542](https://github.com/nominal-io/instro/pull/542) — borrowed OpcUaNodeId conversion
4. **[#526](https://github.com/nominal-io/instro/pull/526) — node type references and NodeId-keyed streaming (this PR)**

Review this PR against [#542](https://github.com/nominal-io/instro/pull/542). Its diff now contains only the final CON-2884 layer, with the supporting API changes available in the preceding reviews.

## Intent
Expose each browsed node's type definition and carry only node identity through read, poll, and subscription APIs.

## Changes
- add type_definition to OpcUaNode and populate it during browse
- add OpcUaNodeId accessors and borrowed static string support
- preserve qualified browse names in browse output
- key read, poll, and subscription APIs on OpcUaNodeId rather than full OpcUaNode values
- generalize list ownership through IntoList
- keep read_attributes_from_nodes deleted

## Key decisions
- unsupported ExpandedNodeId type definitions fall back to the null node ID for now
- read and streaming APIs no longer retain full browse nodes when only identity is needed
- the upstream open62541 ua module is not re-exported
- unrelated lockfile and terminology-only churn was removed

## Potential pitfalls
- adding a public OpcUaNode field breaks downstream struct literals
- browse_name now includes its namespace prefix
- NodeIdKind::String changes from String to Cow<'static, str>
- OpcUaNodeReadBatch can construct multiple node/attribute pairs, while read_nodes still semantically targets value samples; that boundary needs reviewer attention
- broad browsing in [#540](https://github.com/nominal-io/instro/pull/540) can reach the existing one-million-node safety limit sooner

## Validation
- cargo test -p instro-opcua --all-targets
- cargo clippy --locked -p instro-opcua --all-targets -- -D warnings
- cargo fmt --all -- --check


cartercanedy:
@greptileai

cartercanedy:
@claude review

cartercanedy:
Gonna close/reopen so I can clean up the comment mess

PR #540, opened 2026-09-17
fix(opcua): recurse all node classes when browsing

### LINKED ISSUE #537: bug(opcua): recurse all node classes, not just objects and variables
The original mental model that I had for browsing was that only objects and variables mattered, which is not the case. The client should remove the artificial restriction on node types collected during a browse.

cartercanedy:
## Intent
Remove the client-side restriction that recursively browsed only `Object` and `Variable` nodes.

Closes #537

## Changes
- recurse through every node class returned by the server
- represent `DataType`, `ObjectType`, `VariableType`, and `ReferenceType` explicitly
- update the browse regression test to cover Method and View descendants

## Validation
- cargo fmt --all -- --check
- cargo test -p instro-opcua --all-targets


crates/instro-opcua/src/types.rs
```diff
@@ -863,6 +863,10 @@ pub enum OpcUaNodeClass {
     Variable,
     Method,
     View,
+    DataType,
+    ObjectType,
+    VariableType,
+    ReferenceType,
```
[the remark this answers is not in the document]
cartercanedy: This is fine.

cartercanedy:
@greptileai

cartercanedy:
@greptileai

crates/instro-opcua/src/browse.rs
maxleblang, line 104: Are you worried at all about relying on the defaults of an external package to dictate how we recurse each node type? It might be safer to be explicit about the args we want so we don't start to lose nodes or recurse in a weird way and have to debug it all the way back to `BrowseDescription`'s defaults changing.

gmarino2048, line 160: Could you help me understand a bit more how this method is used? Recursive DFS is setting off my spidey senses, usually if you're doing discovery you want a lazy BFS

maxleblang (approved):
LGTM

crates/instro-opcua/src/browse.rs
cartercanedy: not really, tbh

crates/instro-opcua/src/types.rs
```diff
@@ -863,16 +863,40 @@ pub enum OpcUaNodeClass {
...
+        match raw {
+            ua::NodeClass::OBJECT_U32 => Self::Object,
+            ua::NodeClass::VARIABLE_U32 => Self::Variable,
+            ua::NodeClass::METHOD_U32 => Self::Method,
+            ua::NodeClass::VIEW_U32 => Self::View,
+            ua::NodeClass::DATATYPE_U32 => Self::DataType,
+            ua::NodeClass::OBJECTTYPE_U32 => Self::ObjectType,
+            ua::NodeClass::VARIABLETYPE_U32 => Self::VariableType,
+            ua::NodeClass::REFERENCETYPE_U32 => Self::ReferenceType,
+            other => Self::Other(other),
```
gmarino2048, line 884: I kinda agree with the bot on this one... if we don't support the exhaustive list of nodes, removing the "Other" category and marking as non-exhaustive is a more ergonomic way to force the user to handle an unknown node id

crates/instro-opcua/src/browse.rs
cartercanedy: It's the primary routine that automated node discovery. DFS implemented here for simpler cycle detection, but I'm open to changing. Not super high priority considering that we're likely going to drop this as a public API in the future.

crates/instro-opcua/src/types.rs
cartercanedy: We actually do support all of the spec-defined `NodeClass` variants, this was just a bad attempt at mitigating a leaky abstraction over a C enumeration. Since this is a built-in dtype, we should be rejecting the value when it's detected.
Good catch!

crates/instro-opcua/src/browse.rs
```diff
@@ -122,13 +126,27 @@ impl Browse for OpcUaClient {
                     tracing::warn!(
                         target: "opcua::browse",
                         node_id = ?id,
-                        "skipping reference during browse"
+                        "skipping reference during browseL: invalid node id"
```
gmarino2048:
```suggestion
                        "skipping reference during browse: invalid node id"
```

crates/instro-opcua/src/browse.rs
gmarino2048: I've been burned by stack issues w/recursion before. Especially on a meatier function like this one. If it's low prio feel free to kick it to me and I'll rework with looped BFS and a `HashSet`.

maxleblang (dismissed):

crates/instro-opcua/src/types.rs
```diff
@@ -857,48 +856,61 @@ pub struct OpcUaNode {
...
+    Method = 0x04,
+    ObjectType = 0x08,
+    VariableType = 0x10,
+    ReferenceType = 0x20,
+    DataType = 0x40,
+    View = 0x80,
+}
+
+impl OpcUaNodeClass {
+    pub(crate) fn try_from_raw(raw: u32) -> Result<Self> {
```
gmarino2048: Do we want this to be pub(crate)?

gmarino2048 (dismissed):

crates/instro-opcua/src/browse.rs
maxleblang: @gmarino2048 what would that look like? (just curious and trying to expand my algo and Rust knowledge)

crates/instro-opcua/src/types.rs
cartercanedy: for now, this can be changed to a private impl up the stack

maxleblang (approved):

crates/instro-opcua/src/browse.rs
gmarino2048: Will DM you :)

Issue #556, opened 2026-09-19
feat(lib): simplify instruments to provide concise example for docs


arsenovic:
### What happened?

we need two examples which are as concise as possible to save line-space in docs/README :
* one  that illustrates swapping drivers (HAL) 
* one that illustrates both HAL and publishers. 

currently no instro-instrument is concise enough to accomplish  this. the intruments should be modified to make the example concise.  it should be with a very popular instrument (scope or dmm or psu)

PR #558, opened 2026-09-21
docs: re-organize docs

arsenovic:
Renames the mkdocs API reference site's source directory and updates every path reference across build tooling, CI workflows, and repo docs so links and the GitHub Pages build keep working.

## Summary

_Briefly describe what this PR does and why. If it fixes an existing issue, link it (e.g. `Fixes #123`)._

## Type of change

- [ ] Bug fix (`fix`)
- [ ] New feature (`feat`)
- [ ] Breaking change (`feat!` / `fix!`)
- [ ] Refactor (`refactor`)
- [ ] Documentation (`docs`)
- [ ] Chore / tooling (`chore`)

## Verification

_How did you verify this change works? Provide evidence the reviewer can evaluate without reproducing your setup — e.g. test output, screenshots, logs, repro steps + result. If the change is hardware-specific and you don't have the device, say so._

## Tests

- [ ] Unit tests added or updated
- [ ] Existing tests cover this change
- [ ] No tests — explain why:

## Checklist

- [ ] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (e.g. `feat(driver): add support for Keysight E36300`)
- [ ] I have read [CONTRIBUTING.md](CONTRIBUTING.md)
- [ ] Documentation updated if user-facing behavior changed
- [ ] Code follows the style/conventions of the surrounding code

## Notes for reviewers

_Optional — call out specific files, edge cases, or decisions you'd like eyes on._

PR #559, opened 2026-09-21
feat(dmm): add per-function read shorthands to InstroDMM

nhschwab:
Part of #556.

`InstroDMM` requires a two-step dance — `set_measurement_function(...)` then `read()` — and `read()` raises `ValueError` if you forget the first half. That makes the shortest possible DMM example longer than it needs to be, which is what #556 is about.

## What this adds

Six convenience readers on `InstroDMM`:

- `read_dc_voltage()`
- `read_ac_voltage()`
- `read_dc_current()`
- `read_ac_current()`
- `read_resistance()` (2-wire)
- `read_four_wire_resistance()`

Each selects the function **only when it isn't already active**, then delegates to `read()`. Skipping the redundant selection matters: re-sending the function costs a mode change plus the dummy measurement some drivers issue to commit it, and would republish `set_measurement_function.cmd` on every call.

```python
from instro.dmm import InstroDMM
from instro.dmm.drivers import Agilent34401A  # or Keysight34461A, Keithley2400, SimulatedDMM
from instro.lib.publishers import NominalCorePublisher

dmm = InstroDMM(
    name="myDMM",
    driver=Agilent34401A("ASRL3::INSTR"),
    publishers=[NominalCorePublisher(dataset_rid=DATASET_RID)],
)
with dmm:
    print(dmm.read_dc_voltage().latest)
```

## Scope

HAL-only. `DMMDriverBase` is untouched, so no vendor driver changes and nothing breaking. The wrappers compose the existing `@publish_command` / `@publish_measurement` methods rather than re-implementing them, so published channels are identical to the two-call form and there is no double-publishing.

## Docs

- `examples/dmm/dmm_basic.py` rewritten to the concise form above (regenerated `docs/guides/instrumentation/examples/dmm/dmm_basic.mdx`)
- `docs/guides/instrumentation/dmm.mdx`: new "Single-call reads" section, updated Basic Usage, published-channels table, and method reference
- `docs/guides/instrumentation/dmm-simulator.mdx`: client snippet uses the shorthands

## Tests

Two targeted tests in `tests/dmm/test_instro_dmm.py`: the wrapper selects the function then reads on the right channel, and a repeat call in the same function skips the redundant `set_measurement_function` while a switch to another function does re-send it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arsenovic:
these are great. i would suggest that you remove the changes to the .mdx files in the 
 docs for now, as it will conflict with https://github.com/nominal-io/instro/pull/558

nhschwab:
Fixed in 5658b19e. Both review comments describe the same real race — I reproduced it against the branch before fixing:

```
read_dc_voltage() -> {'ut.two_wire_resistance': [4700.0]}
```

One clarification on the mechanism: the channel and the value stay consistent with each other, because `read()` re-reads `_measurement_config` inside its own lock and dispatches and names the channel from the same (switched) function. The harm is to the caller — you call `read_dc_voltage()` and the returned `Measurement` is ohms. The window is between `_read_function`'s unlocked check and `read()` acquiring `_resource_lock`, the one stretch where no lock is held.

## The fix

**1. `_resource_lock` is now `threading.RLock`.** Holding it across the sequence requires reentrancy, since `set_measurement_function()` and `read()` each acquire it themselves — wrapping them with the plain `Lock` self-deadlocks on the first call (verified). This matches `TransportBase`, which already uses an `RLock` so a shared transport can nest lifecycle calls. The swap only ever permits more, never less: nothing released this lock from a non-owning thread, and other threads are excluded exactly as before.

**2. `_read_function` holds the lock across the check and both calls.**

**3. `set_measurement_function` now assigns `_measurement_config` inside the lock.** Not raised in review, but load-bearing for (2): the assignment previously ran *after* the lock was released, leaving a window where another thread had already reprogrammed the instrument to resistance while `_measurement_config` still read `DC_VOLTAGE`. The sequence would then skip the select and read ohms **onto the `dc_voltage` channel** — value and channel disagreeing, which is strictly worse than the bug being fixed. Pre-existing, but the fix doesn't hold without it. `set_digits` / `set_range` / `set_aperture_*` are untouched; they don't affect read dispatch.

## Tradeoff

The publish decorators run after each inner method's body, so the inner calls' publishes now happen while `_read_function` holds the lock, and `publish()` is a synchronous fan-out. I took that as clearly worth it: it covers one small command (only on an actual function change, given skip-if-already-set) plus one measurement, and any contending thread is itself about to do slower VISA I/O. The alternative — private unlocked internals so publishing stays outside the lock — means bypassing the decorators and hand-rolling their validation, which is a lot of structure for a convenience wrapper. Happy to revisit if you'd rather not hold the lock across publishers.

## Test

`test_read_helper_is_atomic_against_concurrent_function_change` parks a `read_dc_voltage()` thread in the check→read window, has a second thread switch to resistance, and asserts the result still lands on `ut.dc_voltage`. Confirmed it fails without the fix and passes with it:

```
with fix REVERTED:  AssertionError: assert 'ut.dc_voltage' in {'ut.two_wire_resistance': [1.0]}
fix restored:       1 passed
```

`tests/dmm` is green (232 passed), `uv run mypy` and `ruff` clean.

instro/dmm/dmm.py
```diff
@@ -393,6 +395,41 @@ def read(self, **kwargs) -> Measurement:
         channel_suffix = self._measurement_config.function.value.lower()
         return self._package_measurement(channel_suffix, response, timestamp, **kwargs)
 
+    def _read_function(self, function: MeasurementFunction, **kwargs) -> Measurement:
```
maxleblang: Maybe rename to `set_function_and_read` or something more descriptive about the fact that this is also selecting the function

instro/dmm/dmm.py
maxleblang: Unrelated to changes made here but I think there's a race condition with touching `_measurement_config` without holding the lock (same for `set_aperture_seconds`, `set_aperture_nplc`, and `set_range`). We should open an issue about needing to hold the lock in those places.

maxleblang (dismissed):
LGTM. Just some nits

instro/dmm/dmm.py
nhschwab: Agreed, opened #570

maxleblang (dismissed):
LGTM

maxleblang (approved):

